PR #54, opened 2025-10-08
Use tableau representation for `PauliString` and `PauliSum` classes

### LINKED ISSUE #53: Change PauliSum to tableau representation
### Describe the enhancement

The fundamental object of the PauliSum should be the tableau, and not a list of PauliStrings.

### Motivation

Should be faster

### Current behavior

Stores a list of PauliStrings. When a gate is applied it loops through these PauliStrings, and acts each time on them

### Proposed behavior

Store the PauliSum as a tableau, then a gate can be applied by right multiplication of the gate symplectic on the tableau. The list of PauliStrings may still be useful, but this can be moved to a method rather than an attribute.

### Additional context

_No response_

### LINKED ISSUE #58: Mixed Species PauliSum Phase behavior
### Describe the bug

Phases of mixed species PauliSums are not handled correctly (some factors seem to be ignored) when performing operations. Specific examples are multiplication and application of gates

### Steps to reproduce

**Example 1:**
```[python]
P = PauliSum(['x2z2 x0z0'], 
             dimensions=[3, 2], 
             weights=[1], phases=[0])
C = Circuit(dimensions=[3, 2],gates=[S(0,3),S(0,3),H(0,3)])
P = C.act(P)
print(P)
```

**Example 2:**
```[python]
P1 = PauliSum(['x1z1 x0z0'], 
             dimensions=[3, 2], 
             weights=[1], phases=[0]) 
print(P1.H()) 
P2 = PauliSum(['x2z2 x0z0'], 
             dimensions=[3, 2], 
             weights=[1], phases=[0]) 
print(P2) 
print('Product:')
print(P1.H() * P2)
```

### Expected behavior

Example 1 should give phase 8
Example 2 should give phase 8

### Actual behavior

Example 1 gives phase 4
Example 2 gives phase 0

### Environment

_No response_

### Additional context

_No response_

ricott1:
Closes #53.

- [x] Add benchmark test for PauliSum multiplication
- [x] Add check and test for minimal qudit dimension
- [x] Use tableau representation rather than x_exp and z_exp
- [x] Use numpy vector operations rather than python loops to make python go brrrrrr
- [x] Add option to skip sanity checks on PauliString
- [x] Use tableau for gate application
- [x] Close #58 

I get the same number of failed tests (5) as I have on the Acquire branch, so if there is a mistake, it's not covered by tests 😅 

The Hadamard benchmark show a 3x speed improvement.


tests/profile_test.py
f-martini, line 1: I think adding a profiling tool is a great idea! I would not do that in a test suite, though.
Originally, I added the notebook scripts/tools/notebooks/Profiling.ipynb to visualize cProfile reports generated in the corresponding CI step using snakeviz. This notebook was never updated and is probably broken.
Anyway, I realized that profiling is more useful if done locally and outside of the CI. 
I suggest extending the notebook into a proper script that:
- take a method name as input (this should be defined in a proper file [as you did below] and imported)
- profile the method exploiting cProfile  and some python magic like globals()\['my_global_function_name'\]() to call the desired function
- display the results using snakeviz
- [optional] setup a tasks (.vscode\tasks.json) that take the method name as input and run the script

ricott1: I see, good point. I think I'd do that in a separate PR then and keep this small profiler locally. 

I finished most of the changes for the PR (still missing the circuit part), i'm just cleaning it up in separate commits. I benchmarked the PauliSum multiplication, and the tableau representation didn't speed it up. 

However, after profiling, I noticing some bottlenecks due to not using numpy vectorized operations wherever possible. By fixing those, I got ~5x speed up.

f-martini: Amazing gains! 💪🏼😄

src/quaos/core/paulis/pauli_sum.py
```diff
@@ -10,6 +10,70 @@
 PauliOrScalarType = Union[PauliType, ScalarType]
 
 
+class PauliStringList(list):
```
f-martini: Split in a proper file

ricott1:
Should the Pauli class also just use a tableau? I think it would make the whole class architecture more consistent. If you agree, I could just add it quickly here

tests/benchmark_test.py
```diff
@@ -1,7 +1,103 @@
 import pytest
+import numpy as np
+
+from quaos.core.circuits.gates import Hadamard
+from quaos.core.paulis.pauli_string import PauliString
+from quaos.core.paulis.pauli_sum import PauliSum
 
 
 @pytest.mark.benchmark
```
f-martini: We can remove test_main_function now

c-nation (approved):

src/quaos/core/paulis/pauli_sum.py
ricott1: Done

tests/benchmark_test.py
ricott1: Done

f-martini (approved):

Issue #56, opened 2025-10-16
Restructure Gate class to be dimensions independent


ricott1:
### Describe the enhancement

Remove the `dimensions` and `qudit_indices` attributes from the Gate class. 
Modify `act` to take also `qudit_indices` as input. 

### Motivation

In this approach, each Gate becomes a 'fixed', in that it is only defined by its symplectic and could be easily reused. This would make using Gates more ergonomic also when building a circuit (no need to initialize a different Gate everytime).

Furthermore, the `U_symplectic_conjugated` and other properties that depend on `n_qudits = len(qudit_indices)` could be precomputed with some sane defaults, or lazily initialized and stored if needed.

With these changes, PHASE, Hadamard etc. become constants, that the user never needs to initialize, they can just be imported and used as:

```
import PHASE
PHASE.act(pauli_sum, [2])
```

### Current behavior

Currently, `dimensions` is used to:

- run consistency checks --> just skip them;
- build the unitaries --> pass the dimensions explicitly (as it is already possible), to also make it more consistent with `act`;
- build the inverse (via the lcm) --> perform the modulo reduction only when the Gate is applied (using the `pauli_sum.lcm`).

### Additional context

We need to be careful about the phase_vector (which depends on `n_qudits`), as in the PHASE Gate it depends on the dimension of the qudit it acts on (there is a special case if it acts on a qubit). Still, one could just handle this in the `act` function.

ricott1:
Looking a bit more into circuits, I think this change would nicely propagate to circuits as well. Basically circuits culd just be a series of (constant references to) gates and qubit_indices to act upon

ldellant:
If I understand correctly, this will be closed automatically when pr #106 will be merged... Solid :)

PR #61, opened 2025-10-24
Add a common interface for Pauli objects

### LINKED ISSUE #55: Use common interface for Pauli, PauliString, and PauliSum
### Describe the enhancement

Use common interface for all Pauli objects. This interface should be:

- `def tableau(self) -> np.ndarray` -> returns the Pauli tableau
- Optionally x_exp and z_exp getters
- `def dimensions(self) -> np.ndarray` -> returns the dimensions of the qudits of the Pauli
- `def n_qudits(self) -> int` -> returns the number of qudits
- Amend functions to update the tableau. This could be an explicit `amend` function or use the internal `__set/get_item__`
- The interface is also used as a target for Gate.act.

The Pauli classes could still have classmethods for simplified initialization.

### Motivation

This will reduce the code surface and simplify development.
For instance, several functions could be simplified to receive a generic Pauli object with the right interface.

### Current behavior

Currently there is no common interface between Pauli, PauliString and PauliSum.

### LINKED ISSUE #57: Remove reference to Pauli strings from PauliSum class
### Describe the enhancement

Once we use the tableau representation, there should be no need to keep the internal `pauli_strings` attribute in a PauliSum and we can remove it.

### Motivation

This would remove a lot of code around internal _get/set_item and also avoid handling syncronization between the tableau and the pauli_strings attribute, ultimately making the code more readable and less error prone.

If we still want to offer an API to access the list of PauliStrings, we could do so and recalculate the PauliString on the fly from the tableau.

### Current behavior

PauliSum stores the `pauli_strings` property. After #54 is merged tho, this property will not be used anymore.

### Proposed behavior

_No response_

### Additional context

_No response_

ricott1:
Closes #55 and #57. Furthermore, standardize how sanity checks are run.

Sanity checks will be run whenever a Pauli object is created from a classmethod (external factory). Internal operations use the class constructor and skip the sanity checks.

- [x] Create the interface PauliObject
- [x] Refactor Pauli
- [x] Refactor PauliString
- [x] Refactor PauliSum
- [x] Fix tests! 😢 There are still 3 failing tests from before (some were fixed)
- [ ] Fix docstrings


f-martini (commented):
what about -> PauliObject instead of -> Pauli as the return type of the classmethod?

ricott1:
@f-martini I can't reply directly to your comment, but where do you mean specifically to change the return type?

ricott1:
@f-martini @c-nation @RickSimon79221 I think you can start with the review. There is still one failing test, but I think we should fix the unitary in another PR, as it would require some extra refactoring that is not related to Pauli (maybe we could do it together with the Gate and Circuit refactoring from #56)

src/sympleq/core/paulis/pauli_object.py
RickSimon79221, line 1: the 'phases', 'set_phases', 'weights', etc methods currently only do pass, is that correct?

RickSimon79221 (approved):
I am not so super familiar with these parts of the code as most of it was written by charlie, but from what I can tell it seems good

c-nation (approved):

PR #77, opened 2025-11-14
Fix flaky test

ricott1:
## 📝 Description
`test_pauli_sum_phase_invariance_matrix_equivalence` was randomly failing because the value for `n_paulis` could randomly be too large to guarantee unique PauliStrings in `PauliSum.from_random`.

It was a more complex fix than expected. Ultimately I chose to split the implementation depending on the maximum possible number of distinct PauliStrings. If this number is less than a certain threshold, I explicitly create a unique set, otherwise we just instantiate random PauliStrings and cross our fingers. The problem is that even for relatively small numbers like n_qudits = 100, n_paulis=20, the naive approach does not work as the number of possible distinct strings grows exponentially.

- [x] The code passes CI.

### DIFF
--- a/src/sympleq/core/paulis/pauli_sum.py
+++ b/src/sympleq/core/paulis/pauli_sum.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 from typing import overload, TYPE_CHECKING, Union
 import numpy as np
+import math
 import scipy.sparse as sp
 import galois
 from sympleq.core.finite_field_solvers import get_linear_dependencies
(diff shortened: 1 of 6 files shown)


f-martini (dismissed):

src/sympleq/core/paulis/pauli_sum.py
```diff
@@ -220,32 +221,54 @@ def from_random(cls,
...
+
+        # For large max_n_paulis we accept that we may have repetitions,
+        # especially if n_paulis is large.
+        def from_large_population() -> list[PauliString]:
+            return [PauliString.from_random(dimensions, seed) for _ in range(n_paulis)]
+
+        # For small population we ensure unicity.
+        def from_small_population() -> list[PauliString]:
+            indices = np.random.choice(max_n_paulis, size=n_paulis, replace=False)
+            # Use mixed-radix encoding to decode indices into exponent to generate PauliString using from_exponents.
```
RickSimon79221: A similar effect can be achieved with the int_to_base and base_to_int functions already implemented in the code (names should be probably changed at some point to better reflect what they are about). For example base_to_int is already in use in the _to_int() method of PauliString. One could write a _from_int() method that just uses int_to_base to construct the pauli_string from the integer like you did here

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -0,0 +1,106 @@
...
+
+from sympleq.core.paulis import PauliSum, PauliString, Pauli
+
+
+prime_list = [2, 3, 5, 7, 11, 13]
+
+
+class TestPauliSumFactories:
+
+    def test_from_pauli_basic(self):
```
RickSimon79221, line 13: Is this test working? Maybe I am stupid and don't see something, but ps here is a paulistring with two paulis, right? Why can Pauli take this (now supposedly an array of length 4) as an argument?

RickSimon79221 (dismissed):
Don't know how I should do this, but overall it seems good, however I added two comments. One more a small question and the other a suggestion for improvement, but nothing is wrong (i think) so I will still approve

tests/core_tests/paulis_tests/pauli_factory_test.py
ldellant: Hmmm I do not see this test in my version... But I agree. ps has 2 Paulis so `Pauli(ps.tableau, ps.dimensions, ps.weights, ps.phases)` shoud give an error I guess...

ricott1:
@RickSimon79221 thanks for the review, I will take a look tonight. Please don't merge :) In general, no need to approve until everything is addressed

tests/core_tests/paulis_tests/pauli_factory_test.py
ricott1: Yea, it was totally bogus, I just sketched the Pauli tests and then forgot to go back at it. It doesnt raise an error because we don't use the factory, so the sanity checks are skipped. Please check again now

src/sympleq/core/paulis/pauli_sum.py
ricott1: I'm struggling to see exactly how to use it. If I call:

```python
for k in indices:
    bases = int_to_bases(k, dimensions)
```

I obtain a list of length `len(dimensions)` right? But here I need both x and z exponents.

In general I agree it's better to reuse functions that already exists.

ricott1: ah, maybe I can draw two sets indices?

ricott1: Ok, I think I figured it out. I can generate the base using the squared dimensions and then take module and integer division to get two exponents rather than one.

RickSimon79221 (approved):

PR #80, opened 2025-11-18
Tests for Pauli upgraded

ldellant:
## 📝 Description
Upgraded tests to mixed dimensions. Added parameters for more randomized tests

- [x] The code passes CI.

### DIFF
--- a/src/sympleq/core/circuits/known_circuits.py
+++ b/src/sympleq/core/circuits/known_circuits.py
@@ -383,9 +383,8 @@ def to_x(pauli_string: PauliString,
          ignore: int | list[int] | None = None) -> Circuit:
     """Finds a circuit to turn a PauliString to ***X*** where the X is at target_index"""
     # Check all dimensions are equal
-    dims = pauli_string.dimensions
-    if len(set(dims)) != 1:
-        raise ValueError(f"All dimensions of pauli_string must be equal, while got {dims}")
+    if len(set(pauli_string.dimensions)) != 1:
+        raise ValueError(f"All dimensions of pauli_string must be equal, while got {pauli_string.dimensions}")
     ignore, target_index = _validate_inputs(pauli_string, target_index, ignore)
     circuit = Circuit(dimensions=pauli_string.dimensions)
     result = _single_qudit_x(pauli_string, target_index, circuit)
--- a/src/sympleq/core/circuits/known_circuits.py
+++ b/src/sympleq/core/circuits/known_circuits.py
@@ -378,9 +378,14 @@ def _multi_qudit_x(pauli_string, target_index, ignore, circuit):
     raise Exception(f"No circuit found to convert {pauli_string} pauli to X")
 
 
-def to_x(pauli_string: PauliString, target_index: int, ignore: int | list[int] | None = None) -> Circuit:
-    """Finds a circuit to turn a PauliString to ****X*** where the X is at target_index"""
-
+def to_x(pauli_string: PauliString,
+         target_index: int,
+         ignore: int | list[int] | None = None) -> Circuit:
+    """Finds a circuit to turn a PauliString to ***X*** where the X is at target_index"""
+    # Check all dimensions are equal
+    dims = pauli_string.dimensions
+    if len(set(dims)) != 1:
+        raise ValueError(f"All dimensions of pauli_string must be equal, while got {dims}")
     ignore, target_index = _validate_inputs(pauli_string, target_index, ignore)
     circuit = Circuit(dimensions=pauli_string.dimensions)
     result = _single_qudit_x(pauli_string, target_index, circuit)
--- a/tests/core_tests/circuits_tests/circuits_test.py
+++ b/tests/core_tests/circuits_tests/circuits_test.py
@@ -17,8 +17,6 @@ class TestCircuits():
         for _ in range(n_tests):
             dimension_chosen = np.random.choice([2, 3, 5, 7, 11, 13, 17], size=1)[0]
             dims = [dimension_chosen for _ in range(np.random.randint(1, max_qudits))]
-            dims = [2, 3]
-            print("banana")
             target_x = np.random.randint(0, len(dims))
             ps = PauliString.from_random(dims)
             if ps.n_identities() == len(dims):
--- a/tests/core_tests/circuits_tests/circuits_test.py
+++ b/tests/core_tests/circuits_tests/circuits_test.py
@@ -1,9 +1,9 @@
 import numpy as np
 import pytest
+from scipy.sparse import issparse
 from sympleq.core.circuits.known_circuits import to_x, to_ix
 from sympleq.core.circuits import Circuit, SUM, SWAP, Hadamard, PHASE, PauliGate
 from sympleq.core.paulis import PauliSum, PauliString
-from scipy.sparse import issparse
 
 
 class TestCircuits():
--- a/tests/core_tests/circuits_tests/circuits_test.py
+++ b/tests/core_tests/circuits_tests/circuits_test.py
@@ -1,21 +1,29 @@
+import numpy as np
+import pytest
 from sympleq.core.circuits.known_circuits import to_x, to_ix
 from sympleq.core.circuits import Circuit, SUM, SWAP, Hadamard, PHASE, PauliGate
 from sympleq.core.paulis import PauliSum, PauliString
-import numpy as np
 from scipy.sparse import issparse
-import pytest
 
 
 class TestCircuits():
 
+    # TODO: I have tried to generalise this to mixed dimensions, but it seems that the to_x
+    # function does not yet support this properly. If this is OK ignore this
+    # (I added an error when mixed dimensions are inputted for the time being)
     def test_to_x(self, n_tests=500):
-        target_x = 0
+        max_qudits = 25
         list_of_failures = []
         for _ in range(n_tests):
-            ps = PauliString.from_random([3, 3, 3, 3])
-            if ps.n_identities() == 4:
+            dimension_chosen = np.random.choice([2, 3, 5, 7, 11, 13, 17], size=1)[0]
+            dims = [dimension_chosen for _ in range(np.random.randint(1, max_qudits))]
+            dims = [2, 3]
+            print("banana")
+            target_x = np.random.randint(0, len(dims))
+            ps = PauliString.from_random(dims)
+            if ps.n_identities() == len(dims):
                 continue
-            c = to_x(ps, 0)
+            c = to_x(ps, target_x)
             if c.act(ps).x_exp[target_x] == 0 or c.act(ps).z_exp[target_x] != 0:
                 print(f"Failed: {ps} -> {c.act(ps)}")
                 list_of_failures.append(ps)
--- a/tests/core_tests/paulis_tests/pauli_factory_test.py
+++ b/tests/core_tests/paulis_tests/pauli_factory_test.py
@@ -12,7 +12,6 @@ class TestPauliSumFactories:
 
     def test_from_pauli_basic(self):
         pauli = Pauli.from_string('x1z2', 3)
-
         ps = PauliString.from_pauli(pauli)
         assert np.array_equal(ps.tableau, pauli.tableau)
         assert np.array_equal(ps.dimensions, pauli.dimensions)
--- a/tests/core_tests/paulis_tests/pauli_factory_test.py
+++ b/tests/core_tests/paulis_tests/pauli_factory_test.py
@@ -12,6 +12,7 @@ class TestPauliSumFactories:
 
     def test_from_pauli_basic(self):
         pauli = Pauli.from_string('x1z2', 3)
+
         ps = PauliString.from_pauli(pauli)
         assert np.array_equal(ps.tableau, pauli.tableau)
         assert np.array_equal(ps.dimensions, pauli.dimensions)
--- a/tests/core_tests/paulis_tests/pauli_test.py
+++ b/tests/core_tests/paulis_tests/pauli_test.py
@@ -5,34 +5,27 @@ import math
 from sympleq.core.paulis import PauliSum, PauliString, Pauli
 from sympleq.core.paulis.constants import DEFAULT_QUDIT_DIMENSION
 
-# List of prime dimensions to test
-# Choose few for faster testing during development (e.g. until 13)
-# Choose all before PR merging
-prime_list = [2, 3, 5, 7]  # , 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
-# Number of random tests to perform for each (most) test function
-# Set to a lower number for faster testing during development (e.g. 25)
-# Set to a higher number before PR merging (e.g. 1000)
-N_tests = 100
+prime_list = [2, 3, 5, 7, 11] #  , 13, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
+N_tests = 250
 
 
 class TestPaulis:
 
-    def choose_dimensions(self,
-                          max_paulis: int,
-                          prime_available: list[int] = prime_list) -> tuple[list[int], int]:
-        dimensions_to_choose_from = prime_available
+    def choose_dimensions_and_n_paulis(self,
+                                       max_paulis: int,
+                                       prime_available: list[int] = prime_list) -> tuple[list[int], int]:
         dimensions = []
         prod = 1
         while True:
-            d = random.choice(dimensions_to_choose_from)
-            if prod * d >= 10 * np.max(dimensions_to_choose_from) + 1:
+            d = random.choice(prime_available)
+            if prod * d >= 10 * np.max(prime_available) + 1:
                 break
             dimensions.append(d)
             prod *= d
         if not dimensions:
-            dimensions = [random.choice(dimensions_to_choose_from)]
+            dimensions = [random.choice(prime_available)]
 
-        # random number of paulis between 1 and prod(dimensions)^2 - 1
+        # fix n_paulis if too large
         n_paulis = min([int(math.prod(dimensions)**2 - 1), max_paulis])
 
         return dimensions, n_paulis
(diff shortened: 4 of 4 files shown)


tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -1,42 +1,69 @@
...
+            d = random.choice(dimensions_to_choose_from)
+            if prod * d >= 10 * np.max(dimensions_to_choose_from) + 1:
+                break
+            dimensions.append(d)
+            prod *= d
+        if not dimensions:
+            dimensions = [random.choice(dimensions_to_choose_from)]
+
+        # random number of paulis
+        n_paulis = min([int(np.prod(dimensions)**2 - 1), max_paulis])  # random.randint(1, int(max(dimensions)**2) - 1)
```
ricott1: numpy integers have a max value, so `np.prod(dimensions)` can unexpectedly give a negative number at the moment (since we use signed integers for now). I would suggest sticking to python math.prod, as python ints can be arbitrarily large (I should have left a comment in this regard).

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -1,42 +1,69 @@
...
+
+        return dimensions, n_paulis
+
+    def random_pauli_string(self,
```
ricott1: Since we are refactoring these tests, can we remove this function? We should just call `PauliString.from_string` in my opinion. If we want to keep because it returns also the exponents, I would change the name (but we could just get the exponents from the PauliString)

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -1,42 +1,69 @@
...
+                            dim: list[int]) -> tuple[PauliString, list[int], list[int]]:
+        r_list = [np.random.randint(0, d) for d in dim]
+        s_list = [np.random.randint(0, d) for d in dim]
+        pauli_str = " ".join(f"x{r}z{s}" for r, s in zip(r_list, s_list))
```
ricott1: I think this conversion is useless, we could just call PauliString.from_exponents

tests/core_tests/paulis_tests/pauli_test.py
ricott1: Remove unnecessary comment

ricott1: Where is `np.prod(dimensions)**2` coming from btw? Isnt the max number of paulis the product of the squares (not the square of the product)?

ricott1 (changes requested):
First quick review

tests/core_tests/paulis_tests/pauli_test.py
ldellant: Ah no that one was used to avoid problems that are extremely rare anyway... In fact it will never be chosen practically :) good catch about the np.prod

tests/core_tests/paulis_tests/pauli_test.py
ldellant: Yes I agree... I was wondering myself whether to remove it or not to be honest. I kept it because I thought there was a reason to be there...

tests/core_tests/paulis_tests/pauli_test.py
ldellant: Useless is a strong word, but I agree

tests/core_tests/paulis_tests/pauli_test.py
ldellant: Brutal

ricott1: But why do we need to pass max_paulis?

tests/core_tests/paulis_tests/pauli_test.py
ricott1: strong words for strong men

tests/core_tests/paulis_tests/pauli_test.py
ldellant: done

tests/core_tests/paulis_tests/pauli_test.py
ldellant: it disappeared so I assume I fixed it :)

tests/core_tests/paulis_tests/pauli_test.py
ldellant: unnecessary is debatable, butr done :)

src/sympleq/core/circuits/known_circuits.py
```diff
@@ -378,9 +378,14 @@ def _multi_qudit_x(pauli_string, target_index, ignore, circuit):
...
 
-def to_x(pauli_string: PauliString, target_index: int, ignore: int | list[int] | None = None) -> Circuit:
-    """Finds a circuit to turn a PauliString to ****X*** where the X is at target_index"""
-
+def to_x(pauli_string: PauliString,
+         target_index: int,
+         ignore: int | list[int] | None = None) -> Circuit:
+    """Finds a circuit to turn a PauliString to ***X*** where the X is at target_index"""
+    # Check all dimensions are equal
+    dims = pauli_string.dimensions
```
ricott1: I would not add a new variable, since we don't reuse it later, and just call `len(set(pauli_string.dimensions))`

tests/core_tests/circuits_tests/circuits_test.py
```diff
@@ -1,21 +1,27 @@
+import numpy as np
```
ricott1, line 1: external import should be grouped together (just revert to previous)

tests/core_tests/circuits_tests/circuits_test.py
```diff
@@ -1,21 +1,27 @@
...
 from sympleq.core.circuits import Circuit, SUM, SWAP, Hadamard, PHASE, PauliGate
 from sympleq.core.paulis import PauliSum, PauliString
-import numpy as np
 from scipy.sparse import issparse
-import pytest
 
 
 class TestCircuits():
 
+    # TODO: I have tried to generalise this to mixed dimensions, but it seems that the to_x
```
ricott1, line 11: Can we then add another (failing) test when mixed dimensions are given (expecting the test to raise)? In this way, when we fix it, the test will fail and we notice immediately.

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -1,42 +1,63 @@
...
 import pytest
+import math
 from sympleq.core.paulis import PauliSum, PauliString, Pauli
 from sympleq.core.paulis.constants import DEFAULT_QUDIT_DIMENSION
 
-prime_list = [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
+# List of prime dimensions to test
+# Choose few for faster testing during development (e.g. until 13)
+# Choose all before PR merging
+prime_list = [2, 3, 5, 7]  # , 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
```
ricott1: remember to select all of them and remove the comments

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -1,42 +1,63 @@
...
+# Number of random tests to perform for each (most) test function
+# Set to a lower number for faster testing during development (e.g. 25)
+# Set to a higher number before PR merging (e.g. 1000)
```
ricott1: I think we could stick with less even, like 50. The CI already takes a while to complete 😅

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -1,42 +1,63 @@
...
+        while True:
+            d = random.choice(dimensions_to_choose_from)
+            if prod * d >= 10 * np.max(dimensions_to_choose_from) + 1:
+                break
+            dimensions.append(d)
+            prod *= d
+        if not dimensions:
+            dimensions = [random.choice(dimensions_to_choose_from)]
+
+        # random number of paulis between 1 and prod(dimensions)^2 - 1
```
ricott1: Why random? also, why the minus 1?

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -1,42 +1,63 @@
...
 
 class TestPaulis:
 
-    def random_pauli_string(self, dim: int) -> tuple[PauliString, int, int, int, int]:
-        r1 = np.random.randint(0, dim)
-        r2 = np.random.randint(0, dim)
-        s1 = np.random.randint(0, dim)
-        s2 = np.random.randint(0, dim)
-        return PauliString.from_string(f"x{r1}z{s1} x{r2}z{s2}", dimensions=[dim, dim]), r1, r2, s1, s2
+    def choose_dimensions(self,
```
ricott1: Maybe we can find a more descriptive name for the return type (e.g. `get_dimensions_and_n_paulis`? which is terrible I know)

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -153,12 +196,17 @@ def test_pauli_sum_multiplication(self):
...
 
     def test_pauli_sum_tensor_product(self):
-
         for dim in prime_list:
-            for _ in range(100):
-                p_string1, r1, r2, s1, s2 = self.random_pauli_string(dim)
-                p_string2, r12, r22, s12, s22 = self.random_pauli_string(dim)
-                p_string3, r13, r23, s13, s23 = self.random_pauli_string(dim)
+            for _ in range(N_tests):
+                p_string1 = PauliString.from_string(f"x{np.random.randint(0, 2 * dim)}z{np.random.randint(0, 2 * dim)}"
```
ricott1: I would use auxiliary variables for the exponents for consistency with previous tests

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -168,24 +216,39 @@ def test_pauli_sum_tensor_product(self):
...
-            [1, 1]
-        ])
-
-        np.testing.assert_array_equal(sp.tableau, expected_matrix)
+        for _ in range(N_tests):
+            # randomly choose dimensions and number of paulis
+            dimensions, n_paulis = self.choose_dimensions(10, prime_available=prime_list)
+            # choose tableau
+            expected_tableau = np.array([
+                (lambda x, z: x + z)(
```
ricott1, line 236: fancy ;)

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -203,153 +266,259 @@ def test_basic_pauli_relations(self):
...
-                p_string1, r1, r2, s1, s2 = self.random_pauli_string(dimensions)
-                p_string2, r12, r22, s12, s22 = self.random_pauli_string(dimensions)
-                p_string3, r13, r23, s13, s23 = self.random_pauli_string(dimensions)
-
-                random_pauli_sum = PauliSum.from_pauli_strings([p_string1, p_string2])
-                random_pauli_sum2 = PauliSum.from_pauli_strings([p_string3])
-                ps_out = random_pauli_sum + random_pauli_sum2
+        for dim in prime_list:
+            for _ in range(N_tests):
+                p_string1 = PauliString.from_string(f"x{np.random.randint(0, 2 * dim)}z{np.random.randint(0, 2 * dim)}"
```
ricott1: As before, I would use aux variables as before

RickSimon79221 (dismissed):
checked the tests and they seem to work, also most comments seem like they have been resolved but I would leave it to @ricott1 to mark them as resolved

src/sympleq/core/circuits/known_circuits.py
ldellant: Is this something I should address in this PR? I agree with you, but I do not think I changed this...

ldellant: Ah yes I changed this :)

Fixed

tests/core_tests/circuits_tests/circuits_test.py
ldellant: you are really annoying :P

tests/core_tests/circuits_tests/circuits_test.py
ldellant: I haven't gone through CIrcuit tests yet, this PR is only for Pauli (at least this was the intention...)

tests/core_tests/paulis_tests/pauli_test.py
ldellant: No pain no gain

tests/core_tests/paulis_tests/pauli_test.py
ldellant: spotless

tests/core_tests/paulis_tests/pauli_test.py
ldellant: clearly somethin i came up with

tests/core_tests/paulis_tests/pauli_test.py
ldellant: aaaahn

tests/core_tests/paulis_tests/pauli_test.py
ldellant: ok ok

tests/core_tests/paulis_tests/pauli_test.py
ldellant: not random and the -1 to avoid identity

tests/core_tests/circuits_tests/circuits_test.py
ricott1: Sounds good, I raised cause the TestCircuits class was modified

tests/core_tests/paulis_tests/pauli_test.py
ricott1: I meant, why in the comment you say it's random? Maybe we can update the comment.

Should we avoid testing the identity for any specific reason?

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -12,7 +12,6 @@ class TestPauliSumFactories:
 
     def test_from_pauli_basic(self):
         pauli = Pauli.from_string('x1z2', 3)
-
```
ricott1, line 15: I would revert this change to have a cleaner history, but no big deal

tests/core_tests/paulis_tests/pauli_test.py
ldellant: because that's the number we actually want, unless it is too big :)

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -1,42 +1,56 @@
 import numpy as np
 import random
-import math
 import pytest
+import math
 from sympleq.core.paulis import PauliSum, PauliString, Pauli
 from sympleq.core.paulis.constants import DEFAULT_QUDIT_DIMENSION
 
-prime_list = [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
+prime_list = [2, 3, 5, 7, 11] #  , 13, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
```
ricott1: Should we uncomment this? Also, again, do we really gain much by having such a large N_tests? If we want to randomly discover edge cases there are better (fuzzing) methods.

tests/core_tests/paulis_tests/pauli_test.py
ldellant: Yes I changed the comment, not a big deal to have the -1, I included it because if the random generator is written such that the Identity is never produced, there is a risk it gets stuck in an infinite loop... 

Since it is random, if Identity can be produced we most likely will get it anyways :)

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -203,153 +267,265 @@ def test_basic_pauli_relations(self):
...
+            psum.standardise()
+
+            # Expected result: all PauliStrings with default weights and phases
+            expected = PauliSum.from_pauli_strings(pauli_strings, weights=[1 + 0j] * n_paulis, phases=[0] * n_paulis)
+            expected.standardise()
+
+        if isinstance(psum, PauliSum) and isinstance(expected, PauliSum):
+            assert psum == expected, (f"PauliSum addition failed, \n obtained \n{psum}\n expected \n{expected}\n,"
+                                      f"with dimensions {dimensions}")
+        else:
```
ricott1: I find this a bit contrived. I would just assert `isinstance(psum, PauliSum) and isinstance(expected, PauliSum)`, so something like


```diff
assert isinstance(psum, PauliSum) and isinstance(expected, PauliSum), (f"PauliSum used are not of type PauliSum, being them \n{psum}\n and \n{expected}\n,"
                           f"with dimensions {dimensions}")
assert psum == expected, (f"PauliSum addition failed, \n obtained \n{psum}\n expected \n{expected}\n,"
                                      f"with dimensions {dimensions}")
```

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -203,153 +267,265 @@ def test_basic_pauli_relations(self):
...
-        assert ps[[0, 2], [0, 2]] == PauliSum.from_string(['x2z0 x1z1', 'x2z0 x2z0'], weights=[1, 0.5], phases=[0, 1],
-                                                          dimensions=[3, 3])
+        for _ in range(N_tests):
+            # Randomly choose dimensions and number of qudits
+            dimensions, n_paulis = self.choose_dimensions_and_n_paulis(10, prime_available=prime_list)
+            n_qudits = len(dimensions)
+
+            # Generate random PauliStrings
+            pauli_strings = []
+            for _ in range(n_paulis):
```
ricott1, line 422: I'll take this as an example of why I don't understand the -1 in `choose_dimensions_and_n_paulis`. Here we generate random PauliStrings, without guaranteeing unicity and including identity.

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -492,144 +710,140 @@ def test_pauli_sum_is_hermitian(self):
...
+        for _ in range(N_tests):
+            # Symmetry and zero diagonal
+            dims = random.choices(prime_list, k=3)
+            P1 = PauliString.from_exponents(x_exp=[1, 0, 0], z_exp=[0, 0, 0], dimensions=dims)  # X on first qudit
+            P2 = PauliString.from_exponents(x_exp=[0, 1, 0], z_exp=[0, 1, 0], dimensions=dims)  # XZ on second qudit
+            P3 = PauliString.from_exponents(x_exp=[0, 0, 1], z_exp=[0, 0, 0], dimensions=dims)  # X on third qudit
+            P4 = PauliString.from_exponents(x_exp=[0, 0, 0], z_exp=[1, 0, 1],
+                                            dimensions=dims)  # Z on first & third qudit
+            S = PauliSum.from_pauli_strings([P1, P2, P3, P4])
+            print("S", S.tableau.shape)
```
ricott1: I would remove random prints

ricott1 (changes requested):
I still have some comments. In general, the PR is a bit bigger than what I would suggest (i know, coming from me is hilarious), and it takes a bit to carefully go through it.

ricott1 (approved):
good job

PR #81, opened 2025-11-18
Add `PauliSum.from_file` factory

ricott1:
## 📝 Description
Add a classmethod to read a PauliSum from file. We could decide to move this to the PauliObject instead.

- [x] The code passes CI.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,1 +1,0 @@
-TEST_DATA_FOLDER = "./_test_data"
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,1 @@
+TEST_DATA_FOLDER = "./_test_data"
--- a/tests/core_tests/paulis_tests/pauli_factory_test.py
+++ b/tests/core_tests/paulis_tests/pauli_factory_test.py
@@ -136,14 +136,8 @@ class TestPauliSumFactories:
         path = TEST_DATA_FOLDER + "/test_pauli_sum_to_and_from_file"
         dimensions = [2, 3, 5, 7, 11]
 
-        P = PauliSum.from_random(10, dimensions)
-
-        spaces = False
-        P.to_file(path, spaces)
-        P_from_file = PauliSum.from_file(path, dimensions)
-        assert P == P_from_file
-
-        spaces = True
-        P.to_file(path, spaces)
-        P_from_file = PauliSum.from_file(path, dimensions)
-        assert P == P_from_file
+        for _ in range(50):
+            P = PauliSum.from_random(10, dimensions)
+            P.to_file(path)
+            P_from_file = PauliSum.from_file(path, dimensions)
+            assert P == P_from_file
--- a/tests/core_tests/paulis_tests/pauli_factory_test.py
+++ b/tests/core_tests/paulis_tests/pauli_factory_test.py
@@ -1,14 +1,27 @@
 import numpy as np
 import pytest
 import math
+import os
+import shutil
+
 
 from sympleq.core.paulis import PauliSum, PauliString, Pauli
+from tests import TEST_DATA_FOLDER
 
 
 prime_list = [2, 3, 5, 7, 11, 13]
 
 
 class TestPauliSumFactories:
+
+    @classmethod
+    def setup_class(cls):
+        if not os.path.exists(TEST_DATA_FOLDER):
+            os.makedirs(TEST_DATA_FOLDER)
+
+    @classmethod
+    def teardown_class(cls):
+        shutil.rmtree(TEST_DATA_FOLDER)
 
     def test_from_pauli_basic(self):
         pauli = Pauli.from_string('x1z2', 3)
@@ -118,3 +131,19 @@ class TestPauliSumFactories:
             for xe, ze, d in zip(ps.x_exp, ps.z_exp, dims):
                 assert 0 <= xe < d
                 assert 0 <= ze < d
+
+    def test_pauli_sum_to_and_from_file(self):
+        path = TEST_DATA_FOLDER + "/test_pauli_sum_to_and_from_file"
+        dimensions = [2, 3, 5, 7, 11]
+
+        P = PauliSum.from_random(10, dimensions)
+
+        spaces = False
+        P.to_file(path, spaces)
+        P_from_file = PauliSum.from_file(path, dimensions)
+        assert P == P_from_file
+
+        spaces = True
+        P.to_file(path, spaces)
+        P_from_file = PauliSum.from_file(path, dimensions)
+        assert P == P_from_file
--- a/tests/core_tests/paulis_tests/pauli_factory_test.py
+++ b/tests/core_tests/paulis_tests/pauli_factory_test.py
@@ -1,27 +1,14 @@
 import numpy as np
 import pytest
 import math
-import os
-import shutil
-
-
+from pathlib import Path
 from sympleq.core.paulis import PauliSum, PauliString, Pauli
-from tests import TEST_DATA_FOLDER
 
 
 prime_list = [2, 3, 5, 7, 11, 13]
 
 
 class TestPauliSumFactories:
-
-    @classmethod
-    def setup_class(cls):
-        if not os.path.exists(TEST_DATA_FOLDER):
-            os.makedirs(TEST_DATA_FOLDER)
-
-    @classmethod
-    def teardown_class(cls):
-        shutil.rmtree(TEST_DATA_FOLDER)
 
     def test_from_pauli_basic(self):
         pauli = Pauli.from_string('x1z2', 3)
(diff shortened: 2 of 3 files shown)


RickSimon79221 (dismissed):
Looks good to me,  the setup_class and teardown_class I can't really judge, but if it works and Francesco says its fine, its probably good :D

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -1,15 +1,28 @@
...
 
 
 prime_list = [2, 3, 5, 7, 11, 13]
 
 
 class TestPauliSumFactories:
 
+    @classmethod
+    def setup_class(cls):
+        if not os.path.exists(TEST_DATA_FOLDER):
```
f-martini: `if not os.path.exists(TEST_DATA_FOLDER):`
This should be created in tests/__init__ level and be available for every test, it should be added to the the repo .gitignore file

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -1,15 +1,28 @@
...
+            os.makedirs(TEST_DATA_FOLDER)
+
+    @classmethod
+    def teardown_class(cls):
```
f-martini: Given that we need to move the folder creation in __init__.py, the `TestPauliSumFactories` class will not be repsonsible for the removal of TEST_DATA_FOLDER, only of its content (better, of the files generated through the TestPauliSumFactories test itself)

tests/__init__.py
```diff
@@ -0,0 +1 @@
+TEST_DATA_FOLDER = "./_test_data"
```
f-martini:
```suggestion
from pathlib import Path

TEST_DATA_FOLDER = str(Path(__file__ ).resolve().parent  / "_test_data")
```

tests/core_tests/paulis_tests/pauli_factory_test.py
f-martini: Another way is to define a TEST_DATA_FOLDER specific for each test, but I think it is better to have test-generated data in a single place so that it is easy to remove globally if needed (especially if we run them locally and for some reason the teardown fail -> it shouldn't).
 
My concern is with tests that are executed in parallel. The tear_down step may cause problems in the case where the shared TEST_DATA_FOLDER is removed.

ricott1: I see, does this mean that we define a Test parent class (with these classmethods) and inherit from each other Test class?

tests/core_tests/paulis_tests/pauli_factory_test.py
ricott1: Could we just remove the whole folder at the end (in the init as u suggest) and avoid handling each file removal?

tests/__init__.py
ricott1: Should it be `TEST_DATA_FOLDER = str(Path(__name__ ).resolve().parent  / "_test_data")`?

ricott1: Disregard previous comment, I was checking this from the python shell, but your suggestion works fine in the file, fixed

tests/__init__.py
```diff
@@ -0,0 +1,20 @@
+from pathlib import Path
+import os
+import shutil
+
```
f-martini:
```suggestion
BASE_TEST_DATA_FOLDER = Path(__file__).resolve().parent / "_test_data"
```

tests/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+class SympleqTest:
+    TEST_DATA_FOLDER = str(Path(__file__).resolve().parent / "_test_data")
```
f-martini:
```suggestion
```

tests/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+    @classmethod
+    def setup_class(cls):
+        if not os.path.exists(SympleqTest.TEST_DATA_FOLDER):
```
f-martini:
```suggestion
        test_data_folder = BASE_TEST_DATA_FOLDER / cls.__name__ # this should work for subclass to
        test_data_folder.mkdir(parents=True, exist_ok=True)
```

tests/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+            os.makedirs(SympleqTest.TEST_DATA_FOLDER)
```
f-martini:
```suggestion
```

tests/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+    @classmethod
+    def teardown_class(cls):
+        shutil.rmtree(SympleqTest.TEST_DATA_FOLDER)
```
f-martini:
```suggestion
        shutil.rmtree(str(BASE_TEST_DATA_FOLDER / cls.__name__))
```

tests/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+    @classmethod
+    def full_path(cls, filename: str) -> str:
+        return SympleqTest.TEST_DATA_FOLDER + filename
```
f-martini:
```suggestion
        return str(BASE_TEST_DATA_FOLDER / cls.__name__ / filename)
```

tests/__init__.py
```diff
@@ -0,0 +1,20 @@
+from pathlib import Path
+import os
+import shutil
+
+
+class SympleqTest:
```
f-martini: Rename it to **FileWritingTestBase** (or something like this, since it is meant to be extended only by tests that write to disk)

tests/core_tests/paulis_tests/pauli_factory_test.py
f-martini: > I see, does this mean that we define a Test parent class (with these classmethods) and inherit from each other Test class?

I like this idea, I commented your solution with some suggestions to have each tests define its own "area of editing"

tests/core_tests/paulis_tests/pauli_factory_test.py
f-martini: yes exactly, that is the idea for tests-specific folders. For the main" _test_data" folder, it is a bit more tricky. For the moment I would leave it there, empty.
The final result in my opinion should be something like this: given
```
TestClass1(FileWritingTest)
   ...
   
TestClass2(FileWritingTest)
   ...

TestClass3
   ...
```
 Starting from
 **tests
 |- _test_data**
 
 You will generate this tree:
  **tests
 |- _test_data
 . . . |-TestClass1
 . . . | . . |- ... test files ...
 . . . |-TestClass2
 . . . . . . |- ... test files ...**
 
 Then in the teardown you go back to the starting point:
  **tests
 |- _test_data**
 
 Removal of subfolders is handled by each test allowing for parallel runs

ricott1: sounds good. One last thing: should we still empty the _test_data folder explicitly in case someone messes up (@ldellant ?) and writes some file to it?

f-martini: Good idea, I would do it in __init__.py: **clear if exists + add a new empty _test_data folder**. 
Doing it after the tests execution would be the best option, but I do not know now how to do it in a clean way.
I would also add "_test_data/" to the .gitignore file.

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -118,3 +118,13 @@ def test_from_random_decoding_correctness_small_population(self):
             for xe, ze, d in zip(ps.x_exp, ps.z_exp, dims):
                 assert 0 <= xe < d
                 assert 0 <= ze < d
+
+    def test_pauli_sum_to_and_from_file(self):
+        path = SympleqTest.full_path("/test_pauli_sum_to_and_from_file")
```
f-martini:
```suggestion
        path = TestPauliSumFactories.full_path("test_pauli_sum_to_and_from_file")
```

ricott1:
@f-martini the approach with the subfolders fails on CI (and also locally) because the filesystem is not writable. I looked online and found this: https://docs.pytest.org/en/stable/how-to/tmp_path.html

I tried it and it works flawlessly. So maybe we can avoid the all parent class in the init file thing and just use tmp_path as the test argument, WIT?

f-martini:
> @f-martini the approach with the subfolders fails on CI (and also locally) because the filesystem is not writable. I looked online and found this: https://docs.pytest.org/en/stable/how-to/tmp_path.html
> 
> I tried it and it works flawlessly. So maybe we can avoid the all parent class in the init file thing and just use tmp_path as the test argument, WIT?

@ricott1 I tested it locally, and I do not have any errors after applying the last fix I proposed:
        - path = SympleqTest.full_path("/test_pauli_sum_to_and_from_file")
        + path = TestPauliSumFactories.full_path("test_pauli_sum_to_and_from_file")
        
The "/" before the folder name is an error: if we use pathlib.Path to compose the path it causes the final test-dir path to be appended to the filesystem root and not to the _test_data folder. 
        That area is protected, indeed, but there is no reason not to write in our ad-hoc data folder provided that we know what we are doing. 👍🏼 
        
That said, I didn't have the time to test it on the CI, so if the fixture approach can be applied to the superclass and propagate to the subclass, I will go for it (at the end of the day, we are doing the same stuff).
I will give a try to our approach just to to know if it work on the CI (the main benefit in practice is that the test-generated data can be easily explored during a debugging session given that the test folder is not in an esotheric position in the fs, I saw that we sahould be able to set a custom folder also for the test_path fixture though).

ricott1:
@f-martini your fix works for me as well. I'll just wait for your tests then. I would say using the pytest standard way is a bit better because we can write less code, but no strong opinion from my side

f-martini (approved):

RickSimon79221 (approved):

PR #84, opened 2025-11-19
closes #83 Literal option to is close

### LINKED ISSUE #83: Add option "literal" to "is_close" method of PauliObject
### Describe the enhancement

I suggest adding an option "literal" to the "is_close" method of PauliObject.

### Motivation

Currently two PauliSums can be the exact same but the way we represent them is different, thus giving false on the comparison. However, sometimes (not always) one just cares about if the PauliSum is unchanged at all (like with symmetries).
That includes effects from ordering and phases (sometimes the weight includes the phase).

### Current behavior

Currently all comparisons compare weights, phases and paulistrings seperatly and even if the phases could be incorperated into the weights to make them the same, is_close would still return false

### Proposed behavior

I would like to add a toggle "literal" that, if true keeps the current behavior, but if false first standadises both PauliSums.

### Additional context

_No response_

RickSimon79221:
## 📝 Description
Added a toggle to PauliSum.is_close() called "literal" that compares PauliSums to another literally (meaning both phases, weights, and paulistrings have to match not only in there values but also in their order) if true, and if false first standadises them, thus accounting for changes in order and phases that might be already merged with the weights

- [x] The code passes CI.

### DIFF
--- a/src/sympleq/core/paulis/pauli_object.py
+++ b/src/sympleq/core/paulis/pauli_object.py
@@ -224,6 +224,9 @@ class PauliObject(ABC):
             Pauli object to compare against.
         threshold : int, optional
             Number of matching decimal digits required for equality. Default is 10.
+        literal : bool, optional
+            If True, compares objects literally in their current form. If False,
+            accounts for reordering and phases in weights. Default is True.
 
         Returns
         -------
--- a/src/sympleq/core/paulis/pauli_object.py
+++ b/src/sympleq/core/paulis/pauli_object.py
@@ -234,25 +234,25 @@ class PauliObject(ABC):
             True if all tableau entries, weights, phases, and dimensions
             match within tolerance; False otherwise.
         """
-        self_PauliSum = self.copy()
-        other_PauliSum = other_pauli.copy()
-        if not literal:
-            self_PauliSum = self_PauliSum.to_standard_form()
-            other_PauliSum = other_PauliSum.to_standard_form()
-
         if not isinstance(other_pauli, self.__class__):
             return False
 
-        if not np.array_equal(self_PauliSum.tableau, other_PauliSum.tableau):
+        ps1 = self
+        ps2 = other_pauli
+        if not literal:
+            ps1 = ps1.to_standard_form()
+            ps2 = ps2.to_standard_form()
+
+        if not np.all(np.isclose(ps1.weights, ps2.weights, 10**(-threshold))):
             return False
 
-        if not np.all(np.isclose(self_PauliSum.weights, other_PauliSum.weights, 10**(-threshold))):
+        if not np.array_equal(ps1.phases, ps2.phases):
             return False
 
-        if not np.array_equal(self_PauliSum.phases, other_PauliSum.phases):
+        if not np.array_equal(ps1.dimensions, ps2.dimensions):
             return False
 
-        if not np.array_equal(self_PauliSum.dimensions, other_PauliSum.dimensions):
+        if not np.array_equal(ps1.tableau, ps2.tableau):
             return False
 
         return True
--- a/src/sympleq/core/paulis/pauli_object.py
+++ b/src/sympleq/core/paulis/pauli_object.py
@@ -214,7 +214,7 @@ class PauliObject(ABC):
         """
         pass
 
-    def is_close(self, other_pauli: Self, threshold: int = 10) -> bool:
+    def is_close(self, other_pauli: Self, threshold: int = 10, literal: bool = True) -> bool:
         """
         Check whether two Pauli objects are approximately equal.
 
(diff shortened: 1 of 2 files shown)


src/sympleq/core/paulis/pauli_object.py
```diff
@@ -224,26 +224,35 @@ def is_close(self, other_pauli: Self, threshold: int = 10) -> bool:
...
+            If True, compares objects literally in their current form. If False,
+            accounts for reordering and phases in weights. Default is True.
 
         Returns
         -------
         bool
             True if all tableau entries, weights, phases, and dimensions
             match within tolerance; False otherwise.
         """
+        self_PauliSum = self.copy()
```
ricott1: I think we can avoid the copy since `to_standard_form` already copies

src/sympleq/core/paulis/pauli_object.py
```diff
@@ -224,26 +224,35 @@ def is_close(self, other_pauli: Self, threshold: int = 10) -> bool:
...
+        other_PauliSum = other_pauli.copy()
```
ricott1: I would use snake_case for variable names

RickSimon79221: so, self_paulisum, etc.? tried avoiding that because it gives the word unknown line for paulisum and self_pauli_sum seems a litte weird, but what do you think?

ricott1: We can also call them ps1 and ps2 I guess :)

src/sympleq/core/paulis/pauli_object.py
```diff
@@ -224,26 +224,35 @@ def is_close(self, other_pauli: Self, threshold: int = 10) -> bool:
...
+        if not literal:
```
ricott1: I would move this after the isinstance test, which is quick, so we avoid unecessary calls to `to_standard_form`

src/sympleq/core/paulis/pauli_object.py
```diff
@@ -224,26 +224,35 @@ def is_close(self, other_pauli: Self, threshold: int = 10) -> bool:
...
+            self_PauliSum = self_PauliSum.to_standard_form()
+            other_PauliSum = other_PauliSum.to_standard_form()
+
         if not isinstance(other_pauli, self.__class__):
             return False
 
-        if not np.array_equal(self.tableau, other_pauli.tableau):
+        if not np.array_equal(self_PauliSum.tableau, other_PauliSum.tableau):
```
ricott1: and since we are at this, can we move this check to be last, since it is in principle the most costly one?

src/sympleq/core/paulis/pauli_object.py
```diff
@@ -214,7 +214,7 @@ def set_weights(self, new_weights: list[int] | np.ndarray):
         """
         pass
 
-    def is_close(self, other_pauli: Self, threshold: int = 10) -> bool:
+    def is_close(self, other_pauli: Self, threshold: int = 10, literal: bool = True) -> bool:
```
ricott1, line 217: I don't know the answer but I will ask the question :) what do you think is the user expectation from this function in terms of default for the literal value? I mean, I guess you think it is `True`, I just wanted to be sure

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -716,14 +716,41 @@ def test_pauli_sum_phase_invariance_matrix_equivalence(self):
...
+        ps1 = PauliString.from_exponents(x_exp=[1, 0], z_exp=[0, 0], dimensions=dims)  # X on qubit
+        ps2 = PauliString.from_exponents(x_exp=[0, 1], z_exp=[0, 1], dimensions=dims)  # XZ on qutrit
+        ps3 = PauliString.from_exponents(x_exp=[1, 1], z_exp=[0, 1], dimensions=dims)  # X on qubit, XZ on qutrit
+
+        psum1 = PauliSum.from_pauli_strings([ps1, ps2], weights=[1.0, 0.5], phases=[0, 1])
+        psum2 = PauliSum.from_pauli_strings([ps1, ps2], weights=[1.0 + 1e-10, 0.5 - 1e-10], phases=[0, 1])
+        psum3 = PauliSum.from_pauli_strings([ps1, ps3], weights=[1.0, 0.5], phases=[0, 1])
+
+        assert psum1.is_close(psum2, threshold=9)
+        assert not psum1.is_close(psum3, threshold=9)
```
ricott1, line 740: Can we add a test where we compare two pauli sums which are equal if strict mode is off but unequal if it is on?

ricott1 (changes requested):
Finished the review (I was missing the test file). Just some comments on possible improvements in the comparison and some minor comments for the test

src/sympleq/core/paulis/pauli_object.py
RickSimon79221: I think we can discuss this and I don't think there is a right answer, but that was, what it was previously doing and I didn't want to change that for now.

tests/core_tests/paulis_tests/pauli_test.py
RickSimon79221: the assert after that is technically kind of doing that, but I will add a check to make it explicit

src/sympleq/core/paulis/pauli_object.py
ricott1: makes sense

ricott1 (approved):
good job!

ldellant:
ahoy :)

PR #85, opened 2025-11-22
Improve Pauli tests coverage

ricott1:
### Improve coverage of Pauli tests
 
| File               | Old | New | Comments |
|--------------------|-----|-----|----------|
| `pauli.py`              | 86% | 100% |          |
| `pauli_object.py` | 81% | 90%  |          |
| `pauli_string.py`  | 63% | 69%  | nice     |
| `pauli_sum.py`     | 71% | 82%  |          |

### Bugs that were discovered while writing tests

  - [x] Additional sanity checks for Pauli and PauliStrings. Before, it was possible to define Pauli and PauliStrings with arbitrary tableaus.
  - [x] Run sanity checks for some PauliSum classmethods.
  - [x] Check that the tableau is 2 dimensional in PauliObject initializer. 
  - [x] Skip dimensions check when multiplying Pauli by a string
  - [x] Cleanup `PauliString.from_exponents` method (remove option to pass a string as argument for x_exp). 
  - [x] Fix dimensions comparison in `PauliObject.__gt__` 
  - [x] Fix weights setter input type 

### Make pytest fast again
I also have a couple commits to speed up Pauli tests in general. It's just making them run on a smaller dimensions vector. I think this is fine for unit testing, if we want to test on larger spaces/more random run/fuzzing, we could do that as acceptance testing.

I also had to skip a failing test: `test_gate_from_target`.

The updated top 5 running times are

```bash
5.70s call     tests/core_tests/circuits_tests/gates_test.py::TestGates::test_phase_table
4.38s call     tests/graph_utils_test.py::TestGraphUtils::test_find_one_permutation
3.94s call     tests/core_tests/circuits_tests/circuits_test.py::TestCircuits::test_circuit_unitary
2.54s call     tests/core_tests/utils_test.py::TestUtils::test_linear_dependence_detection
2.40s call     tests/core_tests/utils_test.py::TestUtils::test_linear_independence
```

On dev instead we have

```bash
24.76s call    tests/core_tests/paulis_tests/pauli_test.py::TestPaulis::test_pauli_sum_is_hermitian
6.73s call     tests/core_tests/find_symplectic_test.py::TestSymplecticSolver::test_map_single_pauli_string_to_target
5.62s call     tests/core_tests/circuits_tests/gates_test.py::TestGates::test_phase_table
5.45s call     tests/core_tests/paulis_tests/pauli_test.py::TestPaulis::test_pauli_sum_commutation_with_matrix
4.09s call     tests/graph_utils_test.py::TestGraphUtils::test_find_one_permutation
```

- [x] The code passes CI.

### DIFF
--- a/src/sympleq/core/paulis/pauli.py
+++ b/src/sympleq/core/paulis/pauli.py
@@ -274,17 +274,20 @@ class Pauli(PauliObject):
             If operand type is unsupported or dimensions mismatch.
         """
         if isinstance(A, str):
-            return self * Pauli.from_string(A)
+            # When multiplying by a string, we retain the own dimensions,
+            # as otherwise it would only work for the default value.
+            A = Pauli.from_string(A)
+            new_tableau = (self.tableau + A.tableau) % self.lcm
+            return Pauli(new_tableau, dimensions=self.dimensions)
 
         if not isinstance(A, Pauli):
             raise Exception(f"Cannot multiply Pauli with type {type(A)}")
 
         if not np.array_equal(self.dimensions, A.dimensions):
-            raise Exception("To multiply two Paulis, their dimensions"
-                            f" {A.dimensions} and {self.dimensions} must be equal")
+            raise ValueError("To multiply two Paulis, their dimensions"
+                             f" {A.dimensions} and {self.dimensions} must be equal")
 
         new_tableau = (self.tableau + A.tableau) % self.lcm
-
         return Pauli(new_tableau, dimensions=self.dimensions)
 
     def __str__(self) -> str:
--- a/src/sympleq/core/paulis/pauli.py
+++ b/src/sympleq/core/paulis/pauli.py
@@ -232,6 +232,28 @@ class Pauli(PauliObject):
         """
         return self.as_pauli_sum().to_hilbert_space()
 
+    def _sanity_check(self):
+        """
+        Validate internal consistency of the Pauli.
+        The Pauli has the extra constraint of having n_paulis() == n_qudits() == 1
+
+        Raises
+        ------
+        ValueError
+            If tableau, dimensions, or exponents are inconsistent or invalid,
+            if n_paulis() != 1, or if n_qudits() != 1.
+        """
+
+        if self.n_paulis() != 1:
+            raise ValueError(
+                f"Invalid tableau for PauliString. The number of Pauli strings should be 1 (got {self.n_paulis()}).")
+
+        if self.n_qudits() != 1:
+            raise ValueError(
+                f"Invalid tableau for Pauli. The number of qudits should be 1 (got {self.n_qudits()}).")
+
+        super()._sanity_check()
+
     def __mul__(self, A: str | Pauli) -> Pauli:
         """
         Multiply this Pauli by another Pauli or parseable string.
--- a/tests/core_tests/paulis_tests/pauli_factory_test.py
+++ b/tests/core_tests/paulis_tests/pauli_factory_test.py
@@ -33,7 +33,7 @@ class TestPauliSumFactories:
         with pytest.raises(ValueError):
             pauli = Pauli.from_string('x1z2 x2z2', 3)
 
-    def test_from_pauli_strings_single(self):
+    def test_pauli_sum_from_pauli_strings_single(self):
         ps = PauliString.from_exponents([0, 1], [1, 0], [3, 2])
         assert ps.shape() == (1, 2)
 
@@ -42,28 +42,32 @@ class TestPauliSumFactories:
         assert P.weights[0] == 2.0
         assert P.phases[0] == 1
 
-    def test_from_pauli_strings_multiple(self):
+    def test_pauli_sum_from_pauli_strings_multiple(self):
         ps1 = PauliString.from_exponents([0, 1], [1, 2], [5, 3])
         ps2 = PauliString.from_exponents([1, 1], [0, 1], [5, 3])
-        P = PauliSum.from_pauli_strings([ps1, ps2], weights=[1.0, 2.0])
+        with pytest.warns(UserWarning, match="Phases are disregarded"):
+            P = PauliSum.from_pauli_strings([ps1, ps2], weights=[1.0, 2.0], inherit_phases=True, phases=[0, 1])
 
         assert P.tableau.shape == (2, ps1.tableau.shape[1])
         assert np.allclose(P.weights, [1.0, 2.0])
 
-    def test_from_pauli_strings_dimension_mismatch(self):
+    def test_pauli_sum_from_pauli_strings_dimension_mismatch(self):
         ps1 = PauliString.from_exponents([0], [1], [3])
         ps2 = PauliString.from_exponents([0, 1], [1, 0], [2, 2])
         with pytest.raises(ValueError):
-            PauliSum.from_pauli_strings([ps1, ps2])
+            _ = PauliSum.from_pauli_strings([ps1, ps2])
 
-    def test_from_pauli_strings_invalid_input(self):
+    def test_pauli_sum_from_pauli_strings_invalid_input(self):
         dimensions = [3, 3]
         ps1 = PauliSum.from_random(2, dimensions)
         ps2 = PauliString.from_exponents([0, 1], [1, 0], dimensions)
         with pytest.raises(ValueError):
-            PauliSum.from_pauli_strings([ps1, ps2])  # type: ignore
+            _ = PauliSum.from_pauli_strings([ps1, ps2])  # type: ignore
 
-    def test_from_pauli_objects(self):
+        with pytest.raises(ValueError):
+            _ = PauliSum.from_pauli_strings([])
+
+    def test_pauli_sum_from_pauli_objects(self):
         ps = PauliSum.from_random(3, 3, rand_phases=True)
         pauli_objects = [
             Pauli.Xnd(1, 3),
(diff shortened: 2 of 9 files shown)


c-nation (dismissed):

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -229,8 +209,8 @@ def test_symplectic_matrix_single_pauli(self):
 
     def test_symplectic_matrix_multiple_paulis(self):
         for _ in range(N_tests):
-            # randomly choose dimensions and number of paulis
-            dimensions, n_paulis = self.choose_dimensions_and_n_paulis(10, prime_available=prime_list)
+            dimensions = choose_random_dimensions(max_product=30)
+            n_paulis = 10
```
ldellant, line 213: Perhaps also define a N_PAULIS to be a global parameter, such that we can modify at the beginning for all functions?

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -892,8 +869,8 @@ def test_pauli_sum_phase_invariance_matrix_equivalence(self):
         matrix. Verify matrices are equal.
         """
         for _ in range(int(np.ceil(np.sqrt(N_tests)))):
-            # randomly choose dimensions and number of paulis
-            dimensions, n_paulis = self.choose_dimensions_and_n_paulis(25, prime_available=prime_list)
+            dimensions = choose_random_dimensions(max_product=30)
+            n_paulis = 5
```
ldellant, line 875: From my previous comment... Here you could say `n_paulis = int(N_PAULIS / 2) + 1`

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -63,8 +64,14 @@ def test_from_string(self):
 
     def test_from_string_list(self):
         s1 = "x1z0 x0z2"
-        P = PauliSum.from_string([s1], [3, 3])
+        P = PauliSum.from_string([s1], 3)
         assert P.tableau.shape[0] == 1
+        assert P.dimensions.tolist() == [3, 3]
```
ldellant, line 124: Cool! So now if all dimensions are the same we can give a single int `3` and when we ask the dimension it returns a list... pretty neat

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -163,3 +164,46 @@ def test_pauli_sum_to_and_from_file(self, tmp_path: Path):
...
+        assert P.tableau.shape == (1, 2)
+        assert P.weights[0] == 2.0
+        assert P.phases[0] == 1
+
+    def test_pauli_string_from_int_mismatch(self):
+        with pytest.raises(ValueError):
+            _ = PauliString.from_exponents(2, 1, [3, 2])
+
+        with pytest.raises(ValueError):
+            _ = PauliString.from_exponents(2, [1, 2], 3)
```
ldellant, line 235: So this way we get a test error if there is no error right? No need to answer if correct, good to know :)

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -163,3 +164,46 @@ def test_pauli_sum_to_and_from_file(self, tmp_path: Path):
...
+        with pytest.raises(ValueError):
+            tableau = np.asarray([[[0, 0, 1, 1]]], dtype=int)
+            _ = PauliString.from_tableau(tableau)
+
+        with pytest.raises(ValueError):
+            tableau = np.asarray([[0, 0, 1, 1], [1, 1, 0, 1]], dtype=int)
+            _ = Pauli.from_tableau(tableau)
+
+        tableau = [0, 0, 1, 1]
+        _ = PauliString.from_tableau(tableau)
```
ldellant: Maybe move after line 189? It is aesthetically a bit unpleasant to have this one afther the previous two, that are supposed to fail if they do not fail...

src/sympleq/core/paulis/pauli.py
```diff
@@ -274,17 +274,20 @@ def __mul__(self, A: str | Pauli) -> Pauli:
             If operand type is unsupported or dimensions mismatch.
         """
         if isinstance(A, str):
-            return self * Pauli.from_string(A)
+            # When multiplying by a string, we retain the own dimensions,
+            # as otherwise it would only work for the default value.
```
ldellant, line 278: fair

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -207,3 +207,76 @@ def test_pauli_string_from_random(self):
         ps3 = PauliString.from_random([2, 3, 5])
         ps4 = PauliString.from_random([2, 3, 5], 12345)
         assert ps3.shape() == ps4.shape()
+
+    def test_pauli_from_exponents(self):
+        p1 = Pauli.from_exponents(1, 2, 3)
+        p2 = Pauli.from_exponents(1, 1)
+        p3 = Pauli.from_exponents(2)
+        p4 = Pauli.from_exponents()
```
ldellant: would it make sense to use random numbers here instead of 1,2,3 ?

ldellant: To an extent we are assessing only the shape, but...

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -207,3 +207,76 @@ def test_pauli_string_from_random(self):
...
+            _ = Pauli.from_tableau(tableau)
+
+    def test_pauli_str(self):
+        tableau = [0, 1]
+        p = Pauli.from_tableau(tableau)
+        assert f"{p}" == "x0z1"
+
+        tableau = [11, 31]
+        p = Pauli.from_tableau(tableau, 37)
+        assert f"{p}" == "x11z31"
```
ldellant: should we loop with random values? I guess it doesn't really matter...

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -42,28 +42,32 @@ def test_from_pauli_strings_single(self):
...
+            _ = PauliSum.from_pauli_strings([ps1, ps2])
 
-    def test_from_pauli_strings_invalid_input(self):
+    def test_pauli_sum_from_pauli_strings_invalid_input(self):
         dimensions = [3, 3]
         ps1 = PauliSum.from_random(2, dimensions)
         ps2 = PauliString.from_exponents([0, 1], [1, 0], dimensions)
         with pytest.raises(ValueError):
-            PauliSum.from_pauli_strings([ps1, ps2])  # type: ignore
+            _ = PauliSum.from_pauli_strings([ps1, ps2])  # type: ignore
```
ldellant, line 64: This one does not work because one of the input is not a paulisum, right? Weren't we discussing of having 

`from_pauli_object` rather than `from_pauli_strings`? 

Such that this would not yield an error here, and would simplify things... But perhaps I forgot why we eventually decided not to...

tests/core_tests/paulis_tests/pauli_factory_test.py
```diff
@@ -78,20 +82,43 @@ def test_from_pauli_objects(self):
         ps1 = PauliSum.from_string(['x0z1 x1z2 x4z4', 'x0z0 x2z2 x3z0'], [2, 3, 5], weights=[2, 0.5], phases=[1, 0])
         ps2 = PauliSum.from_string(['x0z1 x1z1 x0z4', 'x0z0 x2z1 x2z1', 'x1z0 x2z1 x1z1'],
                                    [2, 3, 5], weights=[2, 3.25, 1j], phases=[0, 11, 2])
+
+        P = PauliSum.from_pauli_objects(ps1, inherit_weights=True, inherit_phases=True)
```
ldellant, line 85: ah here it is! Does this make `from_pauli_strings` redundant?

tests/core_tests/paulis_tests/pauli_factory_test.py
ldellant: see next comment

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -1006,3 +1006,69 @@ def test_pauli_object_sub(self):
...
+        psum2 = PauliSum.from_string(["x0z1 x1z0"])
+        assert psum1 < psum2
+
+        psum1 = PauliSum.from_string(["x1z2 x2z1", "x0z0 x1z2"], [2, 3])
+        psum2 = PauliSum.from_string(["x1z0 x0z1", "x1z0 x1z2"], [2, 3])
+        with pytest.raises(Exception):
+            assert psum1 < psum2
+        with pytest.raises(Exception):
+            assert psum1 > psum2
+
```
ldellant, line 1031: Perhaps also a test when dimension mismatch?

```
        psum1 = PauliSum.from_string(["x1z0 x0z1"], [2, 5])
        psum2 = PauliSum.from_string(["x0z1 x1z0"], [2, 3])
        with pytest.raises(Exception):
            assert psum1 < psum2
        with pytest.raises(Exception):
            assert psum1 > psum2
```

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -1006,3 +1006,69 @@ def test_pauli_object_sub(self):
...
+    def test_pauli_phase_setters(self):
+        p1 = Pauli.from_string("x0z1")
+        p2 = Pauli.from_string("x0z1")
+        p3 = Pauli.from_string("x0z1")
+        p4 = Pauli.from_string("x0z1")
+
+        p1.phases[0] = 1.9
+        p2.set_phases([1])
+        p4.phases = [1]
```
ldellant, line 1046: Is this correct behaviour? Because phases are integers 1.9 -> 1 right? 
Should we warn the user about that?

ldellant: Ah yes I recall the discussion, disregard this comment

ldellant: I could have deleted in the first place...

tests/core_tests/paulis_tests/pauli_test.py
ldellant: ah ok seen it, well done!

ldellant (dismissed):
solid solid

tests/core_tests/paulis_tests/pauli_factory_test.py
ricott1: Yes, we do have that as well now. Here I am checking that it does not work (notice the `with pytest.raises(ValueError):`)

tests/core_tests/paulis_tests/pauli_factory_test.py
ricott1: in a sense, yeah. Ultimately, you can eve just sum pauli strings togehter like psum = Paulistring1 + PauliString2

tests/core_tests/paulis_tests/pauli_test.py
ricott1: I think it should depend on the test actually, since the time it takes varies a lot (also notice this is how it was before this PR)

tests/core_tests/paulis_tests/pauli_test.py
ricott1: see above reply

tests/core_tests/paulis_tests/pauli_factory_test.py
ricott1: ok, added random initialization

tests/core_tests/paulis_tests/pauli_factory_test.py
ricott1: done

tests/core_tests/paulis_tests/pauli_factory_test.py
ricott1: Done

ldellant: really? Point for Luca <3

ldellant (approved):

RickSimon79221 (approved):

PR #86, opened 2025-11-22
Handle phases and weights in PauliObject

ricott1:
## 📝 Description
Handle phases and weights consistently for PauliObjects. This means that phases and weights are also relevant for PauliString and Pauli objects.

- [x] Move phases and weights to PauliObject
- [x] Add setters
- [x] Add `had_equivalent_tableau` function to compare two PauliObjects disregarding phases and weights (not sure about the name)
- [x] Raise error if someone tries to set `_dimensions` or `_lcm`
- [x] Use either underscored or public properties consistently in internal methods
- [x] Update tests   
- [x] (Bonus) Move `__repr__` function to PauliObject

Another approach would be to override the `__eq__` behaviour for PauliString and Pauli.

- [x] The code passes CI.

### DIFF
--- a/src/sympleq/core/paulis/pauli.py
+++ b/src/sympleq/core/paulis/pauli.py
@@ -184,32 +184,6 @@ class Pauli(PauliObject):
         return cls.from_exponents(0, 0, dimension)
 
     @property
-    def phases(self) -> np.ndarray:
-        """
-        Returns the phases associated with the Pauli.
-        For a Pauli operator, this is just the trivial phase.
-
-        Returns
-        -------
-        np.ndarray
-            The phases as a 1d-vector.
-        """
-        return np.asarray([0], dtype=int)
-
-    @property
-    def weights(self) -> np.ndarray:
-        """
-        Returns the weights associated with the Pauli.
-        For a Pauli operator, this is just 1.
-
-        Returns
-        -------
-        np.ndarray
-            The weights as a 1d-vector.
-        """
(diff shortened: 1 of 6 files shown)


RickSimon79221 (dismissed):

ricott1:
Something that is still unaccounted for, is setting any np.ndarray element directly, like `P.weights[0] = 1`. This bypasses the setter, and therefore the validations. This problem is present for `weights`, `phases`, and `dimensions`.

To solve this, I think we need to use an auxiliary class that inherits from np.ndarray and then use the `view` function: https://numpy.org/doc/stable/reference/generated/numpy.ndarray.view.html.

All in all, this will not change the public usage of the PauliObject, but would allow us to ensure that `weights`, `phases`, and `dimensions` are consistent.

ricott1:
After more testing, I think the previous point is already addressed implicitly by numpy. When phases or weights are set, the dimensions are checked to ensure that they are consistent, and the values are converted to match the datatype. So for instance, if one tries to set phases[0] = 2.4, the resulting phase is 2. The only thing is that phases are not guaranteed to be modulo 2*lcm, but this shouldnt be a problem since we take the modulo whenever needed.

We handle the dimensions array by setting it to be read-only, so this can't simply be set.


In summary, I think this is fine as it is now, no need for the auxiliary classes.

c-nation (approved):

RickSimon79221 (approved):

PR #87, opened 2025-11-22
Add `PauliSum.from_pauli_objects` classmethod

ricott1:
## 📝 Description
While reviewing some Pauli tests I found out that we called `PauliSum.from_pauli_strings` and passing a list of Pauli rather than PauliString. In this PR I add the from_pauli_objects classmethod to fix that test.

- [x] Add`PauliSum.from_pauli_objects` classmethod
- [x] Fix test
- [x] Fix `PauliSum.from_pauli_strings` type check
- [x] Implement `__add__` for PauliObject, returning a PauliSum. 

- [x] The code passes CI.


RickSimon79221 (dismissed):

ldellant:
Maybe I arrive late for commenting this (and maybe you have implemented all of this already - will hopefully go through the PR soon, unless you accept it before - on't wait for me!)

Yes, `PauliSum.from_pauli_objects` classmethod is the way to go (I think). Desired feature:

1. Whatever the input (single PauliString, many PauliStrings, PauliString and PauliSum(s)), this method should give a PauliSum wth all PauliStrings as input
2. Should use the same weights and phases as given, unless the user decides to overwrite them with some chosen ones
3. Should give an error if dimensions disagree, or weights/phases used as input (if any) are of wrong sizes

If this is already what the method does ignore me... Otherwise ignore me anyway unless you think makes sense :)

ricott1:
> Maybe I arrive late for commenting this (and maybe you have implemented all of this already - will hopefully go through the PR soon, unless you accept it before - on't wait for me!)
> 
> Yes, `PauliSum.from_pauli_objects` classmethod is the way to go (I think). Desired feature:

Maybe another non-exclusive option is to use the `__add__` method to combine PauliObjects into PauliSums? Currently that is only possible for PauliSums, but it would be somewhat natural to to smtg like

```pythone
ps1 = PauliString.from_random(..)
ps2 = PauliString.from_random(..)
P = ps1 + ps2
assert P is a PauliSum with n_paulis == 2
```

this would allow to implement Hamiltonians directly by summing PauliString terms

ldellant:
Yes that would be a very sweet feature to have back. In fact, this is what originally Charlie did, but then because of a circular import he did not manage to implement it fully. But I think that with the PauliObject class that may not be a problem any more (yay :) ) 

In this case, full support. The `__add__` method should always yield a PauliSum, no matter the input, with all PauliObjects that are given. The only check that it should do is that dimensions are all the same (that is important). 

> > Maybe I arrive late for commenting this (and maybe you have implemented all of this already - will hopefully go through the PR soon, unless you accept it before - on't wait for me!)
> > Yes, `PauliSum.from_pauli_objects` classmethod is the way to go (I think). Desired feature:
> 
> Maybe another non-exclusive option is to use the `__add__` method to combine PauliObjects into PauliSums? Currently that is only possible for PauliSums, but it would be somewhat natural to to smtg like
> 
> ```
> ps1 = PauliString.from_random(..)
> ps2 = PauliString.from_random(..)
> P = ps1 + ps2
> assert P is a PauliSum with n_paulis == 2
> ```
> 
> this would allow to implement Hamiltonians directly by summing PauliString terms

RickSimon79221 (approved):

PR #93, opened 2025-11-30
Python (only) publishing workflow

### LINKED ISSUE #39: Continuous Deployment (CD) for quaos library
### Describe the enhancement

As we move toward the first beta version of the framework, it would be nice to have a continuous deployment (CD) GitHub action in place, allowing us to publish the project on PyPI.

### Motivation

Having a project on PyPl allows user to install it and use it for their project in a simple and straightforward way.

### Current behavior

...

### Proposed behavior

...

### Additional context

...

f-martini:
Closes #39
Added publishing workflow for package publication on PYPI.
Updated windows and linux tasks scripts.
[Cleaned commit history]


RickSimon79221 (dismissed):
Not familiar with that part of the code. But at least what concerns me hasn't been changed, so I give the approval here so your workflow is not dependend on waiting for a review

pyproject.toml
```diff
@@ -0,0 +1,40 @@
...
+requires-python = ">=3.11"
+
+[tool.setuptools.packages.find]
+where = ["src"]
+
+[tool.setuptools_scm]
+
+[tool.pytest.ini_options]
+testpaths = ["tests"]
+python_files = ["*_test.py", "*_tests.py", "test_*.py"]
```
ricott1: out of curiosity, do we need to spcify these or are these the defaults?

.vscode/tasks.json
```diff
@@ -19,24 +19,6 @@
             },
             "problemMatcher": []
         },
-        {
-            "label": "Sync Python Requirements",
```
ricott1, line 23: so we are disabling automatic dependency bumping?

.github/actions/deployment/build-wheel/action.yml
```diff
@@ -0,0 +1,49 @@
...
+        echo "🔨 Building wheel for Python ${{ inputs.python-version }}"
+        python -m build --wheel --outdir dist/
+        echo "✅ Build complete:"
+        ls -la dist/
+
+    - name: Install wheel and test dependencies
+      shell: bash
+      run: |
+        pip install dist/*.whl
+        pip install pytest pytest-cov pytest-benchmark
```
ricott1: where do we use `pytest-cov` and `pytest-benchmark`? In later tests in another action?

.github/actions/deployment/publish-project/action.yml
```diff
@@ -0,0 +1,69 @@
...
+        ls -la artifacts/
+        mkdir -p dist
+        cp artifacts/*.whl dist/ 2>/dev/null || echo "No wheels found"
+        echo "📊 Final wheel count:"
+        ls -la dist/
+
+    - name: Set up Python for publishing
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.12"
```
ricott1: why this specific version and not 3.13? Does it matter?

.github/actions/deployment/publish-project/action.yml
```diff
@@ -0,0 +1,69 @@
+name: "Publish Wheels"
+description: "Download artifacts and publish wheels to PyPI"
```
ricott1, line 2: When are we running this action? Probably we should run it only when we tag a new version right?

.github/actions/deployment/upload-artifacts/action.yml
```diff
@@ -0,0 +1,37 @@
+name: "Upload Wheel Artifacts"
+description: "Upload built wheels as GitHub artifacts"
```
ricott1, line 2: Just for me to get a better grasp of all the actions interactions: where do we use these artifacts?

.github/actions/deployment/version-handling/action.yml
```diff
@@ -0,0 +1,66 @@
+name: "Version Injection"
+description: "Replace dynamic version with static version in pyproject.toml"
```
ricott1, line 2: Can you explain what does this action do and why we need it?

pyproject.toml
f-martini: They are just the default👍🏼

f-martini: These have been removed with the last changes

.vscode/tasks.json
f-martini: Yes, the idea is to avoid setting strict dependency versions but support a range of versions so that users' environments can be more easily customized with unrelated packages (pip should be able to find suitable version of possible shared dependencies).

.github/actions/deployment/build-wheel/action.yml
f-martini: in the next step, to be fair pytest-cov is not really required

f-martini:
```suggestion
        pip install pytest pytest-benchmark
```

.github/actions/deployment/publish-project/action.yml
f-martini: Not really necessary, required only to publish with twine, the version itself is not really important but it is convenient to set to ensure reproducibility

.github/actions/deployment/upload-artifacts/action.yml
f-martini: As discussed on monday these are created during the build step, collected into the wheels folder and then published together

.github/actions/deployment/version-handling/action.yml
f-martini: Required to dynamically set the crrent pulished version base on workflow trigger (release, pre-release, PR in dev)

.github/actions/deployment/publish-project/action.yml
f-martini: We run it when we open a release, targeting PyPi and on PR merging targeting PyPi-test

ricott1: Makes sense.

.github/actions/deployment/build-wheel/action.yml
ricott1: Sounds good to remove if not needed

.github/actions/deployment/version-handling/action.yml
ricott1: This is the one im still not getting. What's the difference between 'static' and 'dynamic' version? Ultimately, are we reading the version from pyproject and using it as a variable somewhere else (I guess to publish the wheel)?

f-martini: The static version is a hardcoded value in the pyproject.toml, while a dynamic version is intended to be provided at build time.
The pyproject.toml specification requires the version field to exist regardless of whether it is a fixed value or dynamically supplied during the wheel build process.
In theory, tools like setuptools-scm can be configured directly through pyproject.toml to generate a dynamic version automatically. However, I wasn’t able to get this working using configuration in pyproject.toml alone, so I created a custom versioning action that uses setuptools-scm to derive the version from tags and inject it at build time.

ricott1 (approved):

PR #94, opened 2025-12-03
Tests

c-nation:
## 📝 Description

gate_decomposition_to_circuit is the main addition
There are some small things added in Scripts that are research oriented and can be ignored.

EDIT: also added the fix to Gate.inv() to this PR

- [x ] The code follows the defined style guide.
- [ x] I have added tests that prove my fix is effective or that my feature works.
- [x ] The code passes CI.


src/sympleq/models/Ising.py
```diff
@@ -113,3 +114,25 @@ def site_index(x, y):
         weights.append(h_x)
 
     return PauliSum.from_pauli_strings(paulis, weights=weights, phases=None)
+
+
+def heuristic_clifford_symmetry(n_spins: int):
```
ricott1, line 119: missing return type

src/sympleq/models/Ising.py
```diff
@@ -113,3 +114,25 @@ def site_index(x, y):
...
+    for i in range(n_spins-2):
+        A[-1 - i, 2 + i] = 1
+
+    F = np.block([[A, B], [C, A]])
+    F_G = Gate('F', list(range(n_spins)), F, [2] * n_spins, np.concatenate([np.zeros(n_spins, dtype=int),
+                                                                            np.ones(n_spins, dtype=int)]))
+    return F_G
+
+
+if __name__ == "__main__":
```
ricott1: I would remove these last 3 lines and move it to a test if needed

src/sympleq/core/graphs/graph_coloring.py
```diff
@@ -0,0 +1,101 @@
...
+            new_color[i] = c
+            if c != color[i]:
+                changed = True
+        color = new_color
+        if not changed:
+            break
+    return color
+
+
+def _color_classes(color: np.ndarray) -> Dict[int, List[int]]:
```
ricott1: I think in other parts of the code we use the native `list`, `tuple` `dict` for typing (not the ones imported from typing

src/sympleq/core/graphs/graph_coloring.py
```diff
@@ -0,0 +1,101 @@
...
+
+def _build_base_partition(
+    S_mod: np.ndarray,
+    p: int,
+    *,
+    coeffs: Optional[np.ndarray],
+    col_invariants: Optional[np.ndarray],
+    max_rounds: int = 10,
+    color_mode: str = "wl",      # "wl" | "coeffs_only" | "none"
+) -> Tuple[np.ndarray, Dict[int, List[int]]]:
```
ricott1: same as above

src/sympleq/core/graphs/graph_coloring.py
```diff
@@ -0,0 +1,101 @@
...
+        classes.setdefault(int(c), []).append(i)
+    for c in classes:
+        classes[c].sort()
+    return classes
+
+
+def _build_base_partition(
+    S_mod: np.ndarray,
+    p: int,
+    *,
```
ricott1: what does the `*` do here?

tests/core_tests/circuits_tests/decomposition_test.py
```diff
@@ -1,42 +1,404 @@
...
-# def test_A_invertible(self):
-#     n_qidits_list = [4, 6, 8]
-#     dimension_list = [2, 3, 5]
+    # ---------- preconditioner tests----------
 
-#     for n_qidits in n_qidits_list:
-#         for dimension in dimension_list:
-#             test_gate = Gate.from_random(n_qidits, dimension)
-#             test_circuit = decompose_symplectic_to_circuit(test_gate.symplectic, dimension)
+    def preconditioner_noop_when_A_invertible_test(self, F, p):
```
ricott1: Here and later: missing parameters type

tests/core_tests/circuits_tests/decomposition_test.py
```diff
@@ -57,7 +57,7 @@ def random_symplectic(self, n: int, p: int, rng: np.random.Generator | None = No
 
     # ---------- preconditioner tests----------
 
-    def preconditioner_noop_when_A_invertible_test(self, F, p):
+    def preconditioner_noop_when_A_invertible_test(self, F: np.ndarray, p: int):
```
ricott1: @f-martini Do test names need to start with `test_xxx`? Is it good practice or not needed?

c-nation: I did not start it with test as then it is run automatically. This is called in test_circuit_decomposition_all -> preconditioner. The structure of these tests could be made to be individually called, which would probably be nicer.

c-nation: Its actually quicker to do this restructure than I expected, I will push a restructured set of tests

tests/core_tests/circuits_tests/decomposition_test.py
c-nation: fixed in commit 'tweaks based on review'

src/sympleq/core/graphs/graph_coloring.py
c-nation: fixed in commit 'tweaks based on review'

src/sympleq/core/graphs/graph_coloring.py
c-nation: fixed in commit 'tweaks based on review'

src/sympleq/core/graphs/graph_coloring.py
c-nation: fixed in commit 'tweaks based on review'

tests/core_tests/circuits_tests/decomposition_test.py
f-martini: @ricott1 yes, it is good practice and required for automatic test detection. As @c-nation said, we agreed to define a support methods as class methods with no prefix; in principle, they can be defined outside the class too. 
I suggest to remove the subfix **_test** though, to make it more clear that is not run automatically but just an utility method.

RickSimon79221 (approved):

PR #97, opened 2025-12-08
Symmetries restructure

c-nation:
## 📝 Description

Symmetries moved to core and tests added. They all pass but more coverage is needed.

- [x ] The code follows the defined style guide.
- [x ] I have added tests that prove my fix is effective or that my feature works.
- [ ] The code passes CI.


src/sympleq/core/circuits/gate_decomposition_to_circuit.py
```diff
@@ -577,51 +579,68 @@ def pauli_correction_gate(F: np.ndarray,
...
+        if np.any(Delta % 2 != 0):
             raise AssertionError("For odd p, Δ must be even for Pauli correction.")
-        Δ_half = ((Δ // 2) % p).astype(int)  # well-defined because Δ is even
+        Delta_half = ((Delta // 2) % p).astype(int)  # well-defined because Δ is even
 
-        # Ω over GF(p): [[0, I], [-I, 0]] (note -I ≡ p-1 mod p)
+        # Omega over GF(p): [[0, I], [-I, 0]] (note -I ≡ p-1 mod p)
         Id = np.eye(n, dtype=int)
         Z = np.zeros((n, n), dtype=int)
         Om = np.block([[Z, Id], [(-Id) % p, Z]]) % p
```
f-martini, line 607: Equations (trivial and not-trivial) can be defined in specific functions so that is easier to re-use them. Evaluate the possibility to add a "math" module where to put them

src/sympleq/core/circuits/circuits.py
```diff
@@ -296,3 +299,47 @@ def cleanup(self):
...
         raise NotImplementedError
+
+    def local_circuit(self, indices: list[int] | np.ndarray) -> 'Circuit':
+        """
+        Returns a Circuit object containing only the gates that act only on the specified qudit indices.
+        """
+        indices = list(indices)
+        index_map = {old: new for new, old in enumerate(indices)}
+
+        def _remap_gate(gate: Gate) -> Gate | None:
```
f-martini, line 310: Personally, I prefer to have non-trivial function defined outside the scope of another parent function, even when the parent function (local_circuit in this case) is the only one using the method.

src/sympleq/core/symmetries/block_decomposition.py
```diff
@@ -21,7 +22,7 @@ def independent_columns(B: np.ndarray, p: int) -> np.ndarray:
 # Mode graph helpers
```
f-martini, line 22: Feeling the need to specify this separator inside a file is a good indicator that the file itself can be split 😄 No worry if there is just a function at the beginning, it helps keep things organized, and if someone at some point needs to add functionalities to an algorithm/class can use the additional file (if appropriate) instead of causing the original  to become bloated with code👍🏼

src/sympleq/core/symmetries/clifford.py
```diff
@@ -0,0 +1,235 @@
...
+        U[n:, :n] = np.eye(n, dtype=int)
+        return np.diag(C.T @ U @ C) % mod
+
+    # Gauge choice for T: default ℓ_T = 0  ⇒  h_T = diag(T^T U T) + ℓ_T
+    if l_T is None:
+        l_T = np.zeros(n2, dtype=int)
+    h_T = (diag_U(T) + l_T) % mod
+
+    # Helper to build the composite phase for given h_S, with h_T fixed above
+    def composite_phase(h_S: np.ndarray) -> np.ndarray:
```
f-martini: Same as above 😉

scripts/experiments/symmetries/examples/heuristic_ising.py
```diff
@@ -0,0 +1,218 @@
+import numpy as np
+from sympleq.core.symmetries.clifford import clifford_phase_decomposition
+from sympleq.core.symmetries.block_decomposition import block_decompose_optimal, block_indexes
+from sympleq.models.Ising import ising_chain_hamiltonian, heuristic_clifford_symmetry
+from sympleq.core.circuits import Gate, Circuit, gate_to_circuit
+from sympleq.core.paulis import PauliSum
+from typing import List, Tuple
+
+
+def _local_U_eig(U_block: np.ndarray, tol: float = 1e-10) -> List[Tuple[complex, np.ndarray]]:
```
ldellant, line 10: why this function is defined with "_" at the beginning? If there is a logic let me know such that the (little) that I do I am consistent :)

ldellant: Also, "eigenpairs" refers to (vals, vec), right? If so, ignore this comment :)

scripts/experiments/symmetries/examples/heuristic_ising.py
```diff
@@ -0,0 +1,218 @@
...
+    for choice in np.array(np.meshgrid(*[range(len(le)) for le in local_eig])).T.reshape(-1, len(local_eig)):
+        lam_u = 1.0 + 0j
+        vec = np.array([1.0], dtype=complex)
+        for block_idx, sel in enumerate(choice):
+            u_val, v = local_eig[block_idx][sel]
+            lam_u *= u_val
+            vec = np.kron(vec, v)
+        vec = vec / np.linalg.norm(vec)
+        global_u.append(lam_u)
+        global_vecs.append(vec)
```
ldellant, line 67: We should be a bit careful here... If I understand correctly, here you are diagonalizing each block within a Clifford symmetry, and taking the tensor product of the resulting eigenvectors. 

This is something we cannot do, as the size of these eigenvectors may become large. As such, we should avoid this approach. I believe that you are aware of that, and the approach we discussed with Rick should avoid this problem. I think that this is a "patch" that you have used temporarily before you merge Rick's code into yours (or vice versa). If this is the case, ignore this comment and well done. If not, let's have a chat on how to solve this issue (and well done in any case :) )

scripts/experiments/symmetries/examples/heuristic_ising.py
```diff
@@ -0,0 +1,218 @@
...
+            vec_global = V @ vecs_h[:, idx]
+            vec_global = vec_global / np.linalg.norm(vec_global)
+            eigvals_U_out.append(lam)
+            eigvals_H_out.append(float(lam_h))
+            eigvecs_out.append(vec_global)
+
+    return eigvals_U_out, eigvals_H_out, eigvecs_out
+
+
+def local_joint_eigensystem_for_block(H_prime: PauliSum,
```
ldellant, line 103: I think that my previous comments (i.e., these functions will be replaced after @RickSimon79221 finalizes the symmetry diagonalizer) applies here and in the following, so I skip these functions for the time being :)

At that point, Rick's code will be moved into the core functions and these will be substituted by calling these right?

Again, if not the case let me know...

src/sympleq/core/circuits/circuits.py
ldellant: Also, this function seems a bit dangerous... Consider the following example. 

Gate1 on 1&2 -> Gate2 on 2&3 -> Gate3 on 2

This function would return Gate1 and Gate2 (it seems). But missing Gate 2, which acts on qubit 2 as well, is dangerous. 

In short, @c-nation if you have written this function I am sure there is a reason. If, developing the code, you realize is not used often (or at all), it might be a good idea to remove it or, as Francesco says, move it outside the core functions, where it is required. 

Final comment, it may be a solution to raise an Error if a situation like the one I highlighted above happens. I.e., make this function such that it returns all gates that act on a subset of qubit IFF there are no other gates that act on these qubits, otherwise throw an error.

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
```diff
@@ -550,17 +550,19 @@ def pauli_gate_for_phase_fix(h_raw: np.ndarray,
...
+    Build a Circuit of PauliGates that fixes the phase vector under right-multiplication:
         h_out = h_body + F^T * (2 Ω v)  (mod 2*lcm)
     Choose v in (Z_p)^{2n} so that h_out == h_target.
 
     For odd p: exact correction.
     For p=2: corrects the even part (the best Pauli can do).
+
+    TODO: extend to mixed-dimension qudits. Can be done simply as there is no entangling.
     """
     import numpy as np
```
ldellant, line 567: move `import numpy as np` with other import at the beginning? Or there is a reason for importing it here? 


OMG I am becoming as annoying as @ricott1 ... :O

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
```diff
@@ -577,51 +579,68 @@ def pauli_correction_gate(F: np.ndarray,
     lcm = int(np.lcm.reduce(dims))
     MOD = 2 * lcm
 
-    Δ = (h_target.astype(int) - h_body.astype(int)) % MOD  # even if realizable by Pauli
+    Delta = (h_target.astype(int) - h_body.astype(int)) % MOD  # even if realizable by Pauli
+
+    def _pauli_gates_from_vectors(x_vec: np.ndarray, z_vec: np.ndarray) -> list[PauliGate]:
```
ldellant, line 584: ndarray already ensure that the type is int? If not, we should specify it

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
```diff
@@ -577,51 +579,68 @@ def pauli_correction_gate(F: np.ndarray,
...
         x = v[:n]
         z = v[n:]
-        pauli = PauliString.from_exponents(x, z, dimensions=dims)
-        return PauliGate(pauli)
+        return Circuit(dims, _pauli_gates_from_vectors(x, z))
 
     else:
         # p == 2: Ω has -I == +I; only even Δ are reachable (gives Δ/2 in {0,1})
-        if np.any(Δ % 2 != 0):
+        if np.any(Delta % 2 != 0):
```
ldellant, line 620: perhaps raise a Warning?

src/sympleq/core/circuits/gates.py
```diff
@@ -398,18 +398,37 @@ def copy(self) -> 'Gate':
...
-        lcm = int(pauli.lcm)
-        symplectic = np.eye(2 * n, dtype=int)
-        phase_vector = (2 * symplectic_form(n, lcm) @ np.concatenate([pauli.x_exp, pauli.z_exp])) % (2 * lcm)
-        super().__init__("Pauli", list(range(n)), symplectic, dimensions=pauli.dimensions, phase_vector=phase_vector)
+        full_n = pauli.n_qudits()
+        indexes = [i for i in range(full_n) if pauli.x_exp[i] != 0 or pauli.z_exp[i] != 0]
+        if len(indexes) == 0:
+            raise ValueError("PauliGate must act non-trivially on at least one qudit.")
+
+        dims_local = pauli.dimensions[indexes]
```
ldellant, line 408: stupid comment, maybe call `dims_local` ->  `local_n`, to be more consistend with `full_n`?

src/sympleq/core/circuits/gates.py
```diff
@@ -398,18 +398,37 @@ def copy(self) -> 'Gate':
...
     def copy(self) -> 'Gate':
-        return PauliGate(self.pauli_string)
+        return PauliGate(self.pauli_string, name=self.name)
 
     def unitary(self, dims=None):
-        if dims is None:
-            dims = self.dimensions
-        return pauli_unitary_from_tableau(dims[0], self.pauli_string.x_exp, self.pauli_string.z_exp)
+        # Build the full tensor product over all qudits, placing identities on untouched sites.
+        dims_full = self.pauli_string.dimensions if dims is None else np.asarray(dims, dtype=int)
```
ldellant, line 423: same comment as above... I think that we should try to use the same names for the same variables throughout the framework for clarity. Absolutely do not change all variable names now, just keep it in mind for the future :) 

(By the way, this comment applies to myself as well)

src/sympleq/core/paulis/pauli_sum.py
```diff
@@ -769,26 +769,39 @@ def combine_equivalent_paulis(self):
...
-            for j in range(i + 1, self.n_paulis()):
-                ps2 = self.select_pauli_string(j)
-                if ps1 == ps2:
-                    # FIXME: can overflow for very large n_paulis.
-                    #        One solution could be to normalize it by dividing by the smallest weight.
-                    self._weights[i] = self.weights[i] + self.weights[j]
-                    to_delete.append(j)
-        self._delete_paulis(to_delete)
+        if self.n_paulis() == 0:
+            return
```
ldellant, line 773: `return` what? Not following here but most likely because I am missing something... Maybe raise an error instead of return?

src/sympleq/core/symmetries/block_decomposition.py
```diff
@@ -157,50 +161,79 @@ def symplectic_basis_from_span(B: np.ndarray, p: int) -> np.ndarray:
...
-    U, V = [], []
-    for i in range(0, dS, 2):
-        # Find u such that not in current span and has a nonzero pairing
-        for j in range(dS):
-            if j in idx_used:
-                continue
-            u = S[:, j:j + 1]
-            for k in range(j + 1, dS):
-                if k in idx_used:
+    def try_build(S_perm: np.ndarray) -> np.ndarray | None:
```
ldellant, line 164: maybe I am having troubles with my current viewer, but is `try_build` defined inside `_apply_mode_permutation_to_ST`? Is possible to define it outside?

src/sympleq/core/symmetries/block_decomposition.py
```diff
@@ -157,50 +161,79 @@ def symplectic_basis_from_span(B: np.ndarray, p: int) -> np.ndarray:
...
-        # Orthogonalize v against all previous U, V
-        for u_prev, v_prev in zip(U, V):
-            coeff_u = _scalar(v.T @ Omega @ v_prev % p)
-            coeff_v = _scalar(v.T @ Omega @ u_prev % p)
-            if coeff_u:
-                v = mod_p(v - coeff_u * u_prev, p)
-            if coeff_v:
-                v = mod_p(v + coeff_v * v_prev, p)
+    rng = np.random.default_rng(2025)
+    for _ in range(256):
```
ldellant, line 213: I am always a bit scared in having numbers as `256` or `2025` or `512` into a code, cause I tend to forget about those and there may be applications that need changing them. Maybe define variables at the beginning of the function or initialize these as function variables with these set values, such that the user knows - or, realistically, can find out more easily - what is happening

src/sympleq/core/symmetries/clifford.py
```diff
@@ -0,0 +1,235 @@
...
+        ts = T_gate.symplectic
+        tis = T_gate.inv().symplectic
+        ss = S_gate.symplectic
+        fs = g.symplectic
+        C = Circuit(pauli_sum.dimensions, [T_gate.inv(), S_gate, T_gate])
+        F = C.composite_gate()
+        assert F.act(pauli_sum).to_standard_form() == pauli_sum.to_standard_form()
+
+        assert np.all(F.symplectic == g.symplectic), f'symplectic failed\n{F.symplectic}\n{g.symplectic}'
+        assert np.all(F.phase_vector == g.phase_vector), f'phase vector failed\n{F.phase_vector}\n{g.phase_vector}'
```
ldellant: I think that now that we have the testing framework, we may move all testing there... Up to different opinions from you, or @ricott1 and @f-martini

src/sympleq/core/symmetries/clifford.py
```diff
@@ -0,0 +1,235 @@
...
+
+        assert np.array_equal(fs, (ts @ ss @ tis) % 2), f'symplectic failed\n{fs}\n{ts @ ss @ ts.T}'
+        assert lhs == rhs, f'Localisation failed\n{lhs.__str__()}\n{rhs.__str__()}'
+    return g, S_gate, T_gate
+
+
+def multiple_clifford_symmetries(pauli_sum: PauliSum,
+                                 n_symmetries: int = 1,
+                                 check_symmetry: bool = True,
```
ldellant: same here, `check_symmetry` may be removed and transferred in the testing framework

ldellant (commented):
Very nice! I think that there are tweaks to be done (I also did not manage to go through all code, so I may have missed things), but this PR is very important as it contains the symmetry project, one of the hallmarks of sympleq, which is a lot of work. 

My suggestion is to polish it a bit (so for all others please review the code, at least high level), make sure to have some comprehensive tests that mitigate for errors, and then keep polishing in the next iterations collectively, to enhance performance and user-friendlyness. 

Of course, I let the expert say the last words :)

PR #102, opened 2026-01-15
Add randomized mirror benchmark tool

ricott1:
## 📝 Description
Add a new measurement tool to perform Randomized Mirror Benchmarking.

- [x] The code passes CI.


src/sympleq/core/measurement/noise_model.py
```diff
@@ -0,0 +1,335 @@
...
+
+class DephasingNoise(NoiseModel):
+    def __init__(self, error_rate: float) -> None:
+        # The dephasing channel has tow Klaus operators: the identity and Z.
+        # A single parameter p0 models the noise probability,
+        # representing the probability of having no error.
+        # Kraus operators:
+        # K0 = sqrt(p0) 1 ⊗ 1
+        # K1 = sqrt(1 − p0) Z ⊗ Z
+        if error_rate > 1.0 or error_rate < 0.0:
```
ldellant: abs < 1

src/sympleq/core/measurement/noise_model.py
```diff
@@ -0,0 +1,335 @@
...
+        return [(1.0, PauliSum.from_tableau(tableau, dimensions))]
+
+
+class DephasingNoise(NoiseModel):
+    def __init__(self, error_rate: float) -> None:
+        # The dephasing channel has tow Klaus operators: the identity and Z.
+        # A single parameter p0 models the noise probability,
+        # representing the probability of having no error.
+        # Kraus operators:
+        # K0 = sqrt(p0) 1 ⊗ 1
```
ldellant: Why K0 and K1 are tensor products of two operators?

ldellant: I mean it is a comment so no big deal...

src/sympleq/core/measurement/noise_model.py
```diff
@@ -0,0 +1,335 @@
...
+        return 2
+
+    def kraus_operator_probabilities(self, n_qudits: int) -> list[float]:
+        sqrt_p0 = self.p0
+        sqrt_p1 = (1.0 - self.p0)
+
+        p0_pows = [sqrt_p0 ** i for i in range(n_qudits + 1)]
+        p1_pows = [sqrt_p1 ** i for i in range(n_qudits + 1)]
+
+        output = []
```
ldellant: Ok so it seems to me that here you define a generalized dephasing channel for multiple qudits. I think it is worth doing it for the dephasing and the depolarizing channels, since they are so well studied... 

I wonder whether this increase the complexity of the code. In fact, one can define dephasing channel on a single qubit and apply multiple dephasing channels instead of one combined. (Maybe I did not explain myself well before)

ldellant: One combined dephasing is equivalent to multiple individual dephasing

src/sympleq/core/measurement/noise_model.py
```diff
@@ -0,0 +1,335 @@
...
+        b_x = kraus_tableau[:, :n]
+        b_z = kraus_tableau[:, n:]
+
+        # Compute acquired phases via symplectic form
+        factors = (pauli.lcm // pauli.dimensions)
+        left_acquired_phases = 2 * factors * a_z  @ b_x.T
+        right_acquired_phases = 2 * factors * b_z  @ a_x.T
+
+        # Combine with existing phases and flatten
+        new_phases = (p1 + p2 + p2 + left_acquired_phases + right_acquired_phases) % (2 * pauli.lcm)
```
ldellant: Should we define a function "new_phases", perhaps in pauli_utils, that does exactly this? Because fixing the phase is something we do in more places, if I am not forgetting something. This way we minimize the error opportunities...

src/sympleq/core/measurement/noise_model.py
```diff
@@ -0,0 +1,335 @@
...
+        new_phases = new_phases.reshape(-1)
+
+        # Multiplication between PauliString corresponds to summing the tableaus
+        new_tableau = np.asarray([ps1 + 2 * ps2 for ps1 in pauli.tableau for ps2 in kraus_tableau], dtype=int)
+
+        return PauliSum(new_tableau, pauli.dimensions, pauli.weights, new_phases)
```
ldellant: related to the previous comment, don't we have a method - multiplication - between a PauliSum and a PauliString, that does this? Maybe it was in a previous version of the code? It may be worth having it defined as a method for the PauliSum: multiplication between a PauliSum and another Paulisum with only one element...

src/sympleq/core/measurement/noise_model.py
```diff
@@ -0,0 +1,335 @@
...
+            coefficient = 1
+            for idx, exp in zip(qudit_indices, comb):
+                p_tableau[n_qudits + idx] = exp
+                coefficient *= np.sqrt(self.p0) if exp == 0 else np.sqrt(1.0 - self.p0)
+            output.append((coefficient, PauliSum.from_tableau(p_tableau, dimensions)))
+
+        return output
+
+
+class DepolarizingNoise(NoiseModel):
```
ldellant: General comment. For now, it is absolutely fine, so this specific comment may become an issue for the future, and more generally can be ignored if I got something wrong. 

It seems to me that the channels are, at the moment, more complicated than they could be. I would define a _generic_ noise model in a similar fashion as the DepolarizingNoise or DephasingNoise, with two inputs: 

- [\alpha_0, ..., \alpha_n] : list of coefficients of Kraus operators
- [K_0, ..., K_n] list of Kraus operators, that we choose (for now) to be PauliStrings.

Notice that this is less general than Eq. (3), cause Kraus operators can be whatever they want (more or less), but practically all noise channels employed have this form and is an excellent starting point.

With these two inputs, the only thing needed is now doing pretty much the same thing you do, with an important exception. Instead of applying the channel on _all_ qudits, we only apply it to a chosen _subset_ of qudits (user specified), where the number of these must be the same as the number of qudits on which the Kraus operators act (Identities shall be tensor product with the Kraus to apply the channel to more qudits).

So, to define the Depolarizing channel, we only have to call this generic noise model with inputs
- [\sqrt{p0}, \sqrt{1-p0]
- [I, Z]

and apply this channel to the, e.g., 15th qudit of a PauliSum object.

Makes semi-sense?

src/sympleq/applications/randomized_benchmarking/RMB.py
```diff
@@ -0,0 +1,366 @@
...
+                if p_idx == q_idx:
+                    pauli_string += "x0z1"
+                else:
+                    pauli_string += "x0z0"
+            pauli_strings.append(pauli_string)
+
+        self.initial_state = PauliSum.from_string(pauli_strings, dimensions)
+
+        if random_initial_state:
+            random_phases = self.rng.choice(a=[0, 2], size=self.initial_state.n_paulis())
```
ldellant: 2 * self.rng.choice([0,self.lcm])

(per quando avremo i qudits...

src/sympleq/core/measurement/RMB.py
```diff
@@ -0,0 +1,366 @@
...
+                    if self.rng.choice(a=[False, True]):
+                        # Remove gate from mirrored circuit (right part)
+                        # idx can have max value len(circuit.gates) - 1 == len(self.circuit.gates) / 2 - 1
+                        self.circuit.remove_gate(len(self.circuit) - 1 - idx)
+                    else:
+                        # Remove gate from base circuit (left part)
+                        self.circuit.remove_gate(idx)
+
+        # Initialize noise models probabilities.
+        # In principle, we don't know if there are gates with n_quidits larger than 2,
```
ldellant: in practice unlikely... A long term project would be generalizing everything (not only the RBM) with a choice of Clifford basis. But this is beyond the current scope...

ldellant: For reference, it would be cool for more democratic benchmarking, where the native gates differ...

src/sympleq/applications/randomized_benchmarking/RMB.py
```diff
@@ -0,0 +1,366 @@
...
+            The resulting PauliSum after applying the noisy circuit.
+        """
+        if n_runs < 1:
+            raise ValueError(f"Number of runs must be greater equal to 1 (got {n_runs}).")
+
+        output = self.act(self.initial_state)
+        for _ in range(n_runs - 1):
+            output += self.act(self.initial_state)
+
+        output = output / n_runs
```
ldellant, line 212: isn't this obvious? Shouldn't this be simply the initial state always?

ldellant: ah no ok it takes as argument the circuit and whether it has noise, if I understand correctly...

src/sympleq/applications/randomized_benchmarking/RMB.py
```diff
@@ -0,0 +1,366 @@
...
+
+        return output
```
ldellant, line 214: here we should return errors as well

src/sympleq/core/measurement/noise_model.py
ricott1: here error_rate is really the error probability tho, I don't think it can be negative

src/sympleq/core/measurement/noise_model.py
ricott1: It was a note I took to construct the multi qudit operators, it's fixed now

src/sympleq/core/measurement/noise_model.py
ricott1: Yes, but this is true only for uncorrelated noise, I think I explain this in the docstring at the beginning. In general, in the future we can handle also correlated noise by defining a suitable process matrix

src/sympleq/core/measurement/noise_model.py
ricott1: I removed this function

src/sympleq/core/measurement/noise_model.py
ricott1: This makes sense, and it's pretty much how I planned to implement the factory for a custom channel. I think this handles also the most generic version of kraus operators

src/sympleq/core/measurement/RMB.py
ricott1: Plus I thought it could be useful if we compile a circuit to a single gate

src/sympleq/applications/randomized_benchmarking/RMB.py
ricott1: What do you mean exactly?

src/sympleq/core/measurement/noise_model.py
ldellant: Maybe have an error if it is negative?

src/sympleq/applications/randomized_benchmarking/RMB.py
ldellant: that this is an average (also, it is a sample average), but depending on how many events you are averaging over and on which results we got, each result will have a different confidence/statistical error associated to it. We should calculate those and return them as well

src/sympleq/applications/randomized_benchmarking/RMB.py
```diff
@@ -0,0 +1,363 @@
...
+                    if self.rng.choice(a=[False, True]):
+                        # Remove gate from mirrored circuit (right part)
+                        # idx can have max value len(circuit.gates) - 1 == len(self.circuit.gates) / 2 - 1
+                        self._circuit.remove_gate(len(self._circuit) - 1 - idx)
+                    else:
+                        # Remove gate from base circuit (left part)
+                        self._circuit.remove_gate(idx)
+
+    @classmethod
+    def initial_state(cls,
```
ldellant, line 56: Issue #112 will include a method that does this within the PauliSum class (I will deal with that)

src/sympleq/applications/randomized_benchmarking/noise_model.py
```diff
@@ -0,0 +1,306 @@
...
+    def kraus_probabilities(self) -> list[float]:
+        """
+        Returns the Kraus probabilities for the noise channel.
+
+        p_i = sum_j |α_{ij}|², i.e. the squared norm of each Kraus operator's coefficients.
+        The probabilities sum to 1.
+        """
+        pass
+
+    def apply_quantum_trajectory(self, pauli_sum: PauliSum, qudits: tuple[int, ...]) -> PauliSum:
```
ldellant: food for thoughts... It seems that `apply_quantum_trajectory` is defined as a method ion the `NoiseModel`. I could be missing something, and I think that what I am going to say you are already implementing/thinking to implement in the next iteration... 

In the long term, we most likely want to use trajectories not only for the noise, but also for (non-clifford) circuits. As such, it could be useful to have the `apply_quantum_trajectory` as a method in circuit, where we give an input stabilizer state and it does a certain number of trajectories, with/without noise, with/without non-clifford gates. 

Long term though, long term.......

src/sympleq/applications/randomized_benchmarking/noise_model.py
```diff
@@ -0,0 +1,306 @@
...
+            pauli_sum = kraus_gate.act(pauli_sum, qudits)
+        elif kraus_gate.n_qudits == 1:
+            # Act independently on each qudit
+            for qudit in qudits:
+                pauli_sum = kraus_gate.act(pauli_sum, qudit)
+        # FIXME: handle generic case, else:...
+
+        return pauli_sum
+
+    def act_in_hilbert_space(self, rho: sp.csr_matrix,
```
ldellant: same as previous comment... incorporated into `Circuit`, and pretty condifent that is what you are thingkin doing anyway :)

good idea!

src/sympleq/applications/randomized_benchmarking/noise_model.py
```diff
@@ -0,0 +1,306 @@
...
+
+        # E.g.: single qudit noise on single qudit gate
+        cum_prob = 0
+        if n_qudits == kraus_n_qudits:
+            for kraus_gate, probability in zip(self.kraus_gates(), self.kraus_probabilities()):
+                unitary = embed_unitary(kraus_gate.local_unitary(dimension), qudits, dimensions)
+                output_rho = _apply_unitary(output_rho, unitary, probability)
+                cum_prob += probability
+
+            assert np.abs(cum_prob - 1.0) < 10**(-5), f"cum prob {cum_prob}"
```
ldellant: `10 ** -5` -> `10 ** -10`

src/sympleq/applications/randomized_benchmarking/noise_model.py
```diff
@@ -0,0 +1,306 @@
...
+    def kraus_gates(self) -> list[Gate]:
+        return [GATES.Id, GATES.X, GATES.Y, GATES.Z]
+
+    def kraus_probabilities(self) -> list[float]:
+        return self._probabilities
+
+
+class CompositeNoise(NoiseModel):
+    """
+    A noise model that combines multiple noise models by merging their Kraus operators.
```
ldellant: ... merging their Kraus operators in the order specified by the input.

src/sympleq/applications/randomized_benchmarking/noise_model.py
```diff
@@ -0,0 +1,306 @@
...
+        Operators with the same gates have their probability summed and normalized.
+        """
+        gates_probabilities: dict[Gate, float] = {}
+
+        for model in self.noise_models:
+            for gate, probability in zip(model.kraus_gates(), model.kraus_probabilities()):
+                if not compound_identity and gate == GATES.Id:
+                    continue
+
+                # If operator is unity, don't add to the weight.
```
ldellant: maybe it could have been a good check not to do that, to make sure probability is conserved? If we find the probability of the identity, the sum of all probabilities must be one

src/sympleq/applications/randomized_benchmarking/noise_model.py
```diff
@@ -0,0 +1,306 @@
...
+                # The noiseless channel will have a weight equal to 1 - Sum toher_channels
+
+                if gate in gates_probabilities:
+                    gates_probabilities[gate] += probability
+                else:
+                    gates_probabilities[gate] = probability
+
+        # Normalize weights so probabilities (|w|^2) sum to 1
+        norm = sum(gates_probabilities.values())
```
ldellant: hmmmm a bit dangerous... The moment we are combining kraus operators with their own probability, the combined probability is the product of the individual probabilities. Not fully sure it is equivalent tbh. Would it be enough to to change `+=` -> `*=` in line 288?

src/sympleq/applications/randomized_benchmarking/RMB.py
ldellant: Ah yes... I will I will...

src/sympleq/applications/randomized_benchmarking/RMB.py
```diff
@@ -0,0 +1,334 @@
...
+
+        pauli_strings = []
+        for p_idx in range(n_qudits):
+            pauli_string = ""
+            for q_idx in range(n_qudits):
+                if p_idx == q_idx:
+                    pauli_string += "x0z1"
+                else:
+                    pauli_string += "x0z0"
+            pauli_strings.append(pauli_string)
```
ldellant, line 73: perhaps easier to use `PauliSum.from_tableau(M)` and M here is the rectangular matrix with all 0 first and the identity later?

src/sympleq/applications/randomized_benchmarking/RMB.py
```diff
@@ -0,0 +1,334 @@
...
+
+        ps = PauliSum.from_string(pauli_strings, dimensions)
+        if random_phases:
+            ps.set_phases(rng.choice(a=[0, 2], size=ps.n_paulis()))
```
ldellant, line 77: let's get prepared for the qudit case. `2 * rng. integers(0,ps.lcm)`

src/sympleq/applications/randomized_benchmarking/RMB.py
ldellant: I think you have already written all the functions for here :) :) :)

src/sympleq/applications/randomized_benchmarking/RMB.py
```diff
@@ -0,0 +1,334 @@
...
+            yield pauli
+
+    def act_in_hilbert_space(self, rho: sp.csr_matrix) -> sp.csr_matrix:
+        for gate, qudits in zip(self.gates, self.qudit_indices):
+            dimension = self.dimensions[qudits[0]]
+            unitary = embed_unitary(gate.local_unitary(dimension), qudits, self.dimensions)
+            rho = np.around(unitary @ rho @ unitary.conjugate().transpose(), 10)
+            rho = self.noise_model.act_in_hilbert_space(rho, qudits, self.dimensions)
+
+        assert np.abs(np.sum(rho.diagonal()) - 1.0) < 10**(-5), f"{np.sum(rho.diagonal())}"
```
ldellant: `np.isclose(np.tr(rho.diagonal()), 1.0)`

src/sympleq/core/circuits/gates.py
```diff
@@ -324,8 +324,24 @@ def full_symplectic(self, qudits: tuple[int, ...] | int, n_qudits: int, p: int)
...
+        if p is None:
+            return F
         return F % p
 
+    def __hash__(self):
+        return hash((
+            self.name,
+            tuple(self.symplectic.flatten().tobytes()),
+            tuple(self._phase_vector.tobytes()),
+        ))
```
ldellant, line 344: is this used? Be careful with `hash`, as it could become an intractably larger number. If you need to identify a gate, it is best to find another way rather than hash, or make it very sure that we never call `hash` for, e.g., a gate with 2000 qubits which would have a hash made of millions of digits...

src/sympleq/core/circuits/gates.py
```diff
@@ -524,6 +540,156 @@ def inverse(self) -> _CZ:
...
+        symplectic = np.eye(2, dtype=int)
+        phase_vector = np.array([0, 0], dtype=int)
+
+        super().__init__("Id", symplectic, phase_vector)
+
+    def local_unitary(self, dimension: int | None = None) -> sp.csr_matrix:
+        if dimension is None:
+            dimension = DEFAULT_QUDIT_DIMENSION
+        d = dimension
+        # X|j⟩ = |j+1 mod d⟩, X^{-1}|j⟩ = |j-1 mod d⟩
```
ldellant, line 567: ?

src/sympleq/core/circuits/gates.py
```diff
@@ -524,6 +540,156 @@ def inverse(self) -> _CZ:
...
+        U = np.eye(d, dtype=complex)
+        return sp.csr_matrix(U)
```
ldellant, line 569: We can start improving a bit using sparse matrices... It is not efficient to start from an array and go to sparse, as you include all the zeros. i suggest this:

```
ones = [1 for _ in range(d)]
return sp.csr_matrix((ones , (ones , ones )))
```

This should be much more efficient

src/sympleq/core/circuits/gates.py
```diff
@@ -524,6 +540,156 @@ def inverse(self) -> _CZ:
...
+
+    def phase_vector(self, dimension: int | None = None) -> np.ndarray:
+        # h = 2 * Ω @ tableau, where Ω = [[0, 1], [-1, 0]]
+        # Ω @ [x, 0] = [0, -x], so h = [0, -2x]
+        x = self._tableau[0]
+        if dimension is not None:
+            return np.array([0, -2 * x], dtype=int) % (2 * dimension)
+        return np.array([0, -2 * x], dtype=int)
+
+    def local_unitary(self, dimension: int | None = None) -> sp.csr_matrix:
```
ldellant, line 606: here there is a bit of redundancy I think (also for identity tbh...) 

you should be able to define all of these using `def xz_mat(d: int, aX: int, aZ: int) -> sp.csr_matrix:` 

and that function is already written to save all zeros :)

:)

tests/core_tests/circuits_tests/circuits_test.py
```diff
@@ -439,6 +439,19 @@ def test_roundtrip_preserves_behavior(self):
 
         assert original_result == restored_result
 
+    def test_circuit_inverse(self):
+        n_gates = np.random.randint(10, 20)
+        n_paulis = np.random.randint(4, 8)
+        dimensions = [2, 3, 5, 7]
```
ldellant, line 445:
```
dimensions = choose_random_dimensions(2500)
n_qudits = len(dimensions)
n_gates = rng.integers(n_qudits, 10 * n_qudits ** 2)
n_paulis = rng.integers(n_qudits, 10 * n_qudits ** 2)
```

tests/application_tests/randomized_benchmarking_tests/noise_model_test.py
```diff
@@ -0,0 +1,242 @@
...
+        probs = model.kraus_probabilities(1)
+
+        assert np.isclose(probs[0], 1.0)
+        assert np.allclose(probs[1:], 0.0)
+
+
+class TestRMB:
+    """Tests for RMB class, moved from RMB.py."""
+
+    @pytest.mark.parametrize("n_qudits", [2, 4, 6])
```
ldellant, line 150: `[2, 3, 4, 5, 6]` 

good to have odd numbers as well...

tests/application_tests/randomized_benchmarking_tests/noise_model_test.py
```diff
@@ -0,0 +1,242 @@
...
+        rho_exact = rmb.rho_exact()
+
+        # Check traces are both 1
+        assert np.isclose(np.trace(rho_exact), 1.0)
+        assert np.isclose(np.trace(rho_average), 1.0)
+
+        # Check convergence: error scales with state space size
+        max_error = np.max(np.abs(rho_exact - rho_average))
+        tolerance = 30 * n_qudits / n_runs
+        assert max_error < tolerance, f"Max error {max_error} exceeds tolerance {tolerance}"
```
ldellant, line 173: now that you have the error for the fidelity, you can generalise this by saying
```
dist_norm = abs(p_0_exact - p_0_average) / delta_p_0
assert dist_norm < 5
assert dist_norm > 0.0001
```

here, p_0 are the probabilities to be in the correct state, delta the error of the averaged one. `<5` means that statistically we should not find it bigger, `>0.0001` statistically should not find smaller. We can chat

tests/application_tests/randomized_benchmarking_tests/noise_model_test.py
ldellant, line 150: `@pytest.mark.parametrize("n_runs", [100, 1000, 10000])`

this way we make sure that the two converge as they should in a statistical sense

ldellant (approved):

PR #106, opened 2026-01-23
Make gates dimension-independent

### LINKED ISSUE #56: Restructure Gate class to be dimensions independent

ricott1:
## 📝 Description
Closes https://github.com/QuAOS-Lab/SympleQ/issues/56

- [x] The code passes CI.


src/sympleq/core/circuits/circuits.py
```diff
@@ -1,156 +1,232 @@
...
                 groups[val].append(i)
             return list(groups.values())
+
+        dimensions = np.asarray(dimensions, dtype=int)
         index_sets = index_lists(dimensions)  # list of lists of indexes for each dimension
-        n_dims = len(index_sets)  # number of different dimensions
+        n_dims = len(index_sets)
+
+        single_qudit_gates = [GATES.H, GATES.S]
+        two_qudit_gates = [GATES.SUM, GATES.SWAP]
```
ldellant: Perhaps have these in the input of the function, such that one can choose the generators of the circuits?

(very low priority, for keeping things simple we can keep it this way... Perhaps even remove the SWAP...)

src/sympleq/core/circuits/circuits.py
```diff
@@ -1,156 +1,232 @@
...
-        for i in range(len(self.gates)):
-            if self.gates[i] != other.gates[i]:
+        if len(self._gates) != len(other._gates):
+            return False
+        for i in range(len(self._gates)):
+            if self._gates[i] is not other._gates[i]:  # Compare by identity for singletons
+                return False
+            if self._qudits[i] != other._qudits[i]:
                 return False
         return True
```
ldellant, line 380: This one I would change a bit, to make it more "powerful". Currently, it is comparing gate-by-gate, but two circuits could yield the same unitary while being different. 

What I would do:
1) check they are both circuits and that dimensions are the same (as you already do)
2) For each circuit, create the corresponding single multi-qudit gate that has the same effect of the circuit
3) compare these two with each other. Their tableaus and phase vectors must be equal for the equality to hold. 

This can be a future PR as well, as of now __eq__ works of course :)

ldellant: composite_gate is the function for this :)

src/sympleq/core/circuits/circuits.py
```diff
@@ -182,99 +258,123 @@ def act_iter(self, pauli: PauliString) -> Generator[PauliString, None, None]:
...
-    def unitary(self):
-        known_unitaries = (H, S, CX, SWAP, CNOT, PauliGate)
-        if not np.all([isinstance(gate, known_unitaries) for gate in self.gates]):
-            print([(gate.name, isinstance(gate, known_unitaries)) for gate in self.gates])
-            raise NotImplementedError("Unitary not implemented for all gates in the circuit.")
+    def full_symplectic(self) -> np.ndarray:
+        """Returns the full symplectic matrix of the composite gate."""
+        return self.composite_gate().symplectic
+
+    def unitary(self) -> sp.csr_matrix:
```
ldellant, line 530: For consistency, should we call to_hilbert ?

src/sympleq/core/circuits/circuits.py
```diff
@@ -182,99 +258,123 @@ def act_iter(self, pauli: PauliString) -> Generator[PauliString, None, None]:
...
+                if not np.all(gate_dims == gate_dims[0]):
+                    raise ValueError(
+                        f"Gate {gate.name} acts on qudits with different dimensions {gate_dims}. "
+                        "Multi-qudit gates require equal dimensions."
+                    )
+                d = int(gate_dims[0])
 
-        return m
+            # Get the gate's local unitary
+            U_local = gate.unitary(d)
```
ldellant: Does this work only with certain gates within the circuit? I think @c-nation has generalized it...

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
```diff
@@ -1,10 +1,44 @@
...
 from sympleq.core.paulis import PauliString
 from collections import deque
 
+# Convenience aliases for singleton gates
+H = GATES.H
+H_inv = GATES.H_inv
+S = GATES.S
+S_inv = GATES.S_inv
+SUM = GATES.SUM
+SWAP = GATES.SWAP
```
ldellant: no inverse?

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
```diff
@@ -1,10 +1,44 @@
...
+
+
+def full_gate_symplectic(gate: Gate, qudits: tuple[int, ...] | int, n_qudits: int, p: int) -> np.ndarray:
```
ldellant: why specify "full"?

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
```diff
@@ -129,34 +170,43 @@ def apply_left_once(F_mat: np.ndarray, g: Gate | tuple[str, int]) -> np.ndarray:
...
-                        C_pre.add_gate(op)
+                        gate, qudits = op
+                        assert isinstance(gate, Gate)
+                        C_pre.add_gate(gate, *qudits)
                 return C_pre
 
             seen.add(key)
             queue.append((F_new, new_ops))
 
     raise RuntimeError("Unable to make A invertible within the depth limit.")
```
ldellant, line 145: We should know what is the depth required?

ldellant: To be honest I am failing a bit seeing what is the exact purpose of this function, so my comment may be stupid...

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
```diff
@@ -624,7 +679,7 @@ def pauli_correction_gate(F: np.ndarray,
         return PauliGate(pauli)
 
 
-def gate_to_circuit(big_gate: "Gate") -> "Circuit":
+def gate_to_circuit(big_gate: "Gate", dimensions: list[int] | np.ndarray) -> "Circuit":
```
ldellant, line 635: It seems like this can be used (if not already used) to get the unitary matrix (see my previous comment in function `unitary` or arbitrary circuits... If not already working :)

src/sympleq/core/circuits/gates.py
```diff
@@ -1,415 +1,622 @@
...
+        dimensional_factor = pauli.lcm // relevant_lcm
         acquired_phases = (linear_terms + quadratic_terms) * dimensional_factor
 
         new_phases = (pauli.phases + acquired_phases) % (2 * pauli.lcm)
 
         return pauli.__class__(tableau=new_tableau, dimensions=pauli.dimensions,
                                weights=pauli.weights, phases=new_phases)
 
-    def copy(self) -> 'Gate':
+    def unitary(self, dimension: int) -> sp.csr_matrix:
```
ldellant: rename to to_hilbert for consistency?

src/sympleq/core/circuits/gates.py
```diff
@@ -1,415 +1,622 @@
...
+        np.ndarray
+            The d^n x d^n unitary matrix, where n is the number of qudits the gate acts on.
         """
-        return Gate(self.name, self.qudit_indices.copy(), self.symplectic.copy(), self.dimensions,
-                    self.phase_vector.copy())
+        # Default implementation for generic gates - subclasses should override
+        raise NotImplementedError(
+            f"unitary() not implemented for {self.__class__.__name__}. "
+            "Override in subclass or use a specific gate class."
+        )
```
ldellant, line 245: Ok, it seemed that with the previous gate decomposition function, this should be "easy". To my knowledge, @c-nation has already done it :)

src/sympleq/core/circuits/gates.py
```diff
@@ -1,415 +1,622 @@
...
+        symplectic = symplectic_random_transvection(n_qudits, dimension, num_transvections)
+        # For random gates, we use zero phase vector (phases depend on specific gate sequence)
+        phase_vector = np.zeros(2 * n_qudits, dtype=int)
 
+        return _GenericGate("random", symplectic, phase_vector)
 
-def _scalar_dim(dim):
-    return int(np.asarray(dim).reshape(-1)[0])
+    @classmethod
+    def solve_from_target(cls, input_tableau: np.ndarray, target_tableau: np.ndarray) -> "Gate":
```
ldellant: Generic comment. I think that building sympleQ around this function is going to make sympleQ quite unique. 

In general, let us aim at making this function very efficient and build other functionalities on this one. E.g., all algorithms that require Cliffords to do "something" may use this one.

src/sympleq/core/circuits/gates.py
```diff
@@ -1,415 +1,622 @@
...
+class _GenericGate(Gate):
+    """
+    A generic gate for computed gates (e.g., from transvection, composite gates).
+    These are not singletons and don't have pre-computed inverses.
+    Uses parent's inverse() which computes the inverse dynamically.
+    """
+    pass
+
+
+class _Hadamard(Gate):
```
ldellant: Maybe capitalized, HADAMARD?

src/sympleq/core/circuits/gates.py
```diff
@@ -1,415 +1,622 @@
...
+        else:
+            symplectic = np.array([
+                [0, -1],   # image of X:  X -> -Z
+                [1, 0]     # image of Z:  Z -> X
+            ], dtype=int)
+            name = "H"
+
+        super().__init__(name, symplectic)
+
+    def unitary(self, dimension: int) -> sp.csr_matrix:
```
ldellant: perhaps here and everywhere to_hilbert?

src/sympleq/core/circuits/gates.py
```diff
@@ -1,415 +1,622 @@
...
+        for j in range(d):
+            for k in range(d):
+                in_idx = j * d + k  # |j,k⟩
+                if self._is_inverse:
+                    out_k = (k - j) % d
+                else:
+                    out_k = (j + k) % d
+                out_idx = j * d + out_k  # |j, out_k⟩
+                U[out_idx, in_idx] = 1.0
+        return sp.csr_matrix(U)
```
ldellant: Low priority: 

maybe define the SUM gate as a composition of CZ and Hadamard (this works for qudits as well I think?)...

This would make the code a bit more compact

src/sympleq/core/circuits/gates.py
```diff
@@ -1,415 +1,622 @@
...
-            dims = self.dimensions
+    @property
+    def CNOT(self) -> _SUM:
+        return self._SUM
 
-        unitary = tensor([S_mat(dims[i]) if i in self.qudit_indices else I_mat(dims[i]) for i in range(len(dims))])
-        return unitary
+    @property
+    def CX(self) -> _SUM:
+        return self._SUM
```
ldellant: Do we need both CNOT and CX and SUM? 

I would remove CNOT, and keep CX and CZ

tests/core_tests/circuits_tests/circuits_test.py
```diff
@@ -84,29 +83,30 @@ def random_pauli_string(self, dim, n_qudits):
     def make_random_circuit(self, n_gates, n_qudits, dimension) -> Circuit:
```
ldellant: don't we have already a function that creates random circuits? Do we have to define one here as well?

tests/core_tests/circuits_tests/gates_test.py
```diff
@@ -22,10 +21,8 @@ def states(dim):
         return ps1, ps2, ps3, ps4, ps5, p_sum
 
     def random_pauli_sum(self, dim, n_paulis=10):
```
ldellant, line 25: don't we have a function that does this already?

ldellant (commented):
There is a lot here, and probably we can improve the code a bit in the future. But for the time being, excellent work everyone, it is a lot here and this is the basis of sympleQ :)

src/sympleq/core/circuits/gates.py
ricott1: I kept them as in the original, but I agree we should simplify

src/sympleq/core/circuits/circuits.py
ricott1: @c-nation @RickSimon79221 opinions?

src/sympleq/core/circuits/gates.py
ricott1: ja, fine for me, here I was keeping the same notation as before

tests/core_tests/circuits_tests/circuits_test.py
ricott1: good point, removed

tests/core_tests/circuits_tests/gates_test.py
ricott1: good point, removed

src/sympleq/core/circuits/circuits.py
ricott1: Maybe we could have a different constructor for that? like `from_gates` or something

src/sympleq/core/circuits/circuits.py
ricott1: but that would be more `__eq__` on the unitary, as you say. I would find it a bit confusing if == returns true for different circuits, even if equivalent as unitaries. Maybe we should have a different function `is_equivalent`?

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
ricott1: for the inverse the idea is that you would call SWAP.inverse()

src/sympleq/core/circuits/circuits.py
ricott1: IF that is fine for you, I would resolve this and include the new constructor only when needed (in a separate PR), since this is already quite big

src/sympleq/applications/measurement/allocation.py
```diff
@@ -517,13 +517,13 @@ def construct_diagnostic_circuits(circuit_list: list[Circuit]) -> list[Circuit]:
...
-        for g in circ.gates:
-            C_diag.add_gate(g.copy())
-            if g.name == 'H':
-                C_diag.add_gate(g.copy())
-                C_diag.add_gate(g.copy())
-                C_diag.add_gate(g.copy())
+        C_diag = Circuit.empty(dimensions=circ.dimensions)
+        for gate, qudits in zip(circ.gates, circ.qudit_indices):
+            C_diag.add_gate(gate, *qudits)
+            if gate.name == 'H':
```
ricott1, line 523: here probably we should use `isinstance`, but im not sure about what we are trying to achieve here (if we don't want inverses, for isntance)

tests/core_tests/circuits_tests/decomposition_test.py
```diff
@@ -169,12 +177,14 @@ def M_block_only(self, n, p, rng=np.random.default_rng(42)):
     def local_Du_synthesis(self, p):
         for u in range(1, p):
             ops = _emit_local_ops_for_D(0, u, p)
-            F = Circuit([p], ops).composite_gate().symplectic % p
+            gates = [op[0] for op in ops]
+            qudits = [op[1] for op in ops]
+            F = Circuit([p], gates, qudits).composite_gate().symplectic % p
```
ricott1: Here and below, we should use Circuit constructors

src/sympleq/core/circuits/known_circuits.py
```diff
@@ -35,48 +40,28 @@ def add_phase(xz_pauli_sum: PauliSum, qudit_index: int, qudit_index_2: int, phas
     if xz_pauli_sum.dimensions[qudit_index] != 2 or xz_pauli_sum.dimensions[qudit_index] != 2:
         raise ValueError("Pauli dimensions must be equal to 2")
 
+    dims = [2 for i in range(xz_pauli_sum.n_qudits())]
+    q1, q2 = qudit_index, qudit_index_2
+
     if phase_key == 'SSSS':
-        C = Circuit(dimensions=[2 for i in range(xz_pauli_sum.n_qudits())])
-        return C
+        return Circuit(dimensions=dims)
```
ricott1: here and below, use constructors

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
ricott1: it was called full_symplectc before as well. I think it refers to the fact that it is relative to the full space (considers all dimensions as a whole rather than being local)

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
ricott1: This function I'm leaving as it is but it should be cleaned up

src/sympleq/applications/measurement/allocation.py
ldellant, line 232: please add 
#TODO: rewrite `diagonalize` using sympleQ core functions, e.g., `decompose_symplectic_to_circuit`

ldellant: see issue #117

src/sympleq/applications/measurement/allocation.py
ldellant: This has to do with ensuring the output is a computational state when employing qudits. Practically, whenever there is a Hadamard, we are copying it to ensure that the output remains a computational state. The idea is that single qudit gates are much more stable than multi-qudit gates...

@RickSimon79221 I see here that we are copying Hadamard three times... This works well even for ququints qusept etc. correct? I think we already discussed about this :P

src/sympleq/applications/measurement/allocation.py
ldellant, line 563: This is one of the functions that could be generalized with the noise model that @ricott1 is currently developing :)

src/sympleq/core/circuits/circuits.py
ldellant: Agree... For (e.g.) noise, it would make a difference...

src/sympleq/models/symmetric_hamiltonian.py
ldellant, line 15: `int_to_bases` should be defined already into core... We should only have one in sympleQ to avoid modifying one and not the other...

ldellant: there is one into `utils.py`, that should either be eliminated or imported here...

tests/core_tests/circuits_tests/circuits_test.py
```diff
@@ -376,31 +320,131 @@ def test_circuit_unitary(self):
...
-                    g = gate(PauliString.from_random(dimensions))
-                    c = Circuit(dimensions, [g])
-                else:
-                    c = Circuit(dimensions, [gate(indices[0], dim)])
-                    g = gate(indices[0], dim)
-
-                U_c = c.unitary().toarray()
-                U_g = g.unitary(dims=dimensions).toarray()
-                assert np.allclose(U_c, U_g)
+        pass
```
ldellant: `pass`? Should we remove this? It seems that the test (seemingly) removed was ok?

tests/core_tests/circuits_tests/gates_test.py
ldellant: isn't it still here ahah?

ldellant (approved):

tests/core_tests/circuits_tests/circuits_test.py
ricott1: Removed as another test already checks this

tests/core_tests/circuits_tests/gates_test.py
ricott1: Now removed for real :)

src/sympleq/core/circuits/gates.py
```diff
@@ -1,415 +1,629 @@
...
+            Local Hilbert space dimension (e.g., 2 for qubits).
+        num_transvections : int | None
+            Number of transvections to compose. If None, defaults to 4*n_qudits.
+
+        Returns
+        -------
+        Gate
+            A random Clifford gate with the generated symplectic matrix.
+        """
+        from sympleq.core.circuits.random_symplectic import symplectic_random_transvection
```
c-nation: Import can move to the top I think

src/sympleq/core/circuits/gates.py
```diff
@@ -1,415 +1,629 @@
...
+        ------
+        ValueError
+            If the tableaus have different shapes or are not mappable via Clifford.
+
+        Notes
+        -----
+        Currently only works for GF(2) (qubits). The input and target must have
+        matching symplectic product matrices for a Clifford mapping to exist.
         """
+        from sympleq.core.circuits.find_symplectic import map_pauli_sum_to_target_tableau
```
c-nation: as above

src/sympleq/core/circuits/gates.py
c-nation: Yes there is a way to do it, but we need a bit of a restructure to have it done here without a circular import. This can be another pr some time

c-nation (approved):

src/sympleq/core/circuits/gates.py
ricott1: done

src/sympleq/core/circuits/gates.py
ricott1: done

tests/core_tests/circuits_tests/gates_test.py
ricott1: actually i had to keep it cause the function constructs a very specific pauli_sum

src/sympleq/core/circuits/circuits.py
ricott1: we now have `to_hilbert_space = unitary` as alias

src/sympleq/core/circuits/circuits.py
ldellant: added issue

RickSimon79221 (approved):

PR #118, opened 2026-02-02
Adding graph automorphism

c-nation:
## 📝 Description

- [x] The code passes CI.


src/sympleq/core/symmetries/clifford.py
```diff
@@ -0,0 +1,95 @@
...
+
+# def _labels_union(independent: list[int], dependencies: dict[int, list[tuple[int, int]]]) -> list[int]:
+#     """Given a set of independent indices and dependencies (mapping from index to list of (index, label) tuples),
+#     return the sorted union of all involved indices.
+#     check if needed - move to utils if so
+# """
+#     return sorted(set(independent) | set(dependencies.keys()))
+
+
+def find_clifford_symmetries(
```
ricott1, line 7: Why do we need this wrapper around `clifford_graph_automorphism_search`? It looks to me that it's only passing over the input parameters

c-nation: It was only really to have something with a more reasonable name in the right place.

src/sympleq/core/graphs/graph_automorphism.py
```diff
@@ -0,0 +1,374 @@
...
+    # Prepared consistency kernel
+    consistency = _ConsistencyChecker(S_mod, use_bitset)
+
+    # Static domain order from base classes (largest first)
+    base_order = sorted(base_classes.keys(), key=lambda c: -len(base_classes[c]))
+    domain_order = [i for c in base_order for i in base_classes[c]]
+
+    # Prepare references for the leaf checks so they dont have to be computed every time
+    pauli_weighted = pauli_sum.copy()
+    pauli_weighted.weight_to_phase()
```
ricott1: Can you give me some intuition of what we are doing here? Why do we need to call wight_to_phase, standardise and then weight_to_phase again?

src/sympleq/core/graphs/graph_automorphism.py
```diff
@@ -0,0 +1,374 @@
...
+    base_weights = np.asarray(pauli_sum.weights)
+    base_phases = np.asarray(pauli_sum.phases, dtype=int)
+    basis_indices = np.asarray(independent_labels, dtype=int)
+    basis_source_ps = pauli_sum[basis_indices]
+
+    dims_array = np.asarray(pauli_sum.dimensions, dtype=int)
+    row_basis_cache: dict[str, np.ndarray] = {}
+    if dims_array.size and np.all(dims_array == dims_array[0]):
+        p_uni = int(dims_array[0])
+        if p_uni == 2:
```
ricott1, line 289:
```suggestion
        cache = _select_row_basis_indices(base_tableau % p_uni, p_uni, base_tableau.shape[1])
        if p_uni == 2:
            row_basis_cache["gf2"] = cache
        else:
            row_basis_cache["gfp"] = cache
```

Find this a tad cleaner

src/sympleq/core/graphs/graph_automorphism.py
```diff
@@ -0,0 +1,374 @@
...
+            rem = sum((not used[y] and (coeffs is None or coeffs[i] == coeffs[y])) for y in base_classes[bi])
+            if rem < best_rem:
+                best_i, best_rem = i, rem
+                if rem <= 1:
+                    break
+        return best_i
+
+    leaf_ctx = _LeafContext(
+        p=p,
+        two_lcm=2 * int(pauli_sum.lcm),
```
ricott1:
```suggestion
        two_lcm=2 * pauli_sum.lcm,
```

src/sympleq/core/symmetries/clifford.py
ricott1: So `clifford_graph_automorphism_search` is meant to be reused somewhere else in the future (even potentially)?

scripts/configs/cspell.json
```diff
@@ -41,6 +42,7 @@
         "eigval",
         "eigvec",
         "elec",
+        "Elim",
```
ricott1, line 45: this is not important by any means, but where is this used?

src/sympleq/core/symmetries/clifford.py
c-nation: Probably not - the structure could change around this. Before I had find_clifford_symmetries doing some work that then moved into the automorphism search function, so find_clifford_symmetries just became a wrapper. we could move the whole automorphism search into clifford.py, or keep the wrapper, or something else. The structure here could be optimised much better I'm sure.

src/sympleq/core/graphs/graph_automorphism.py
c-nation: it is creating two copies, and standardise has phase_to_weight in it. I have pushed a version that removes the need for both copies, though, it was not necessary.

scripts/configs/cspell.json
c-nation: in get_linear_dependencies

RickSimon79221 (approved):

f-martini (approved):

PR #122, opened 2026-02-07
Typing pattern proposal

f-martini:
## 📝 Description

This PR introduces a dedicated typing.py file to centralize shared variant types that would otherwise be duplicated across the codebase, where they tend to drift and become inconsistent over time. Naming these variants makes API intent explicit and improves readability without obscuring supported inputs.

Using formatters (adapter may be a better name) at module boundaries allows us to normalize variants once and work internally with stable, predictable types (as we already do), and avoids repeating isinstance checks and normalization logic throughout the code (we do this as well but there is no infrastructure to aggregate code with similar goals, i.e. the formatters/adapter ).

The main benefit is reduced duplication and safer evolution when supported inputs change, since updates occur in a single place. The trade-off is a small amount of indirection and the fact that type aliases are enforced only by static tooling (which, on the other hand, is nice since it does not impact performance).

- [x] The code passes CI.


ldellant (approved):
Very very good, the fewest place we define the same function/type/etc., the better it is :)

.gitignore
ricott1, line 1: Seems unrelated to the PR?

src/sympleq/core/paulis/typing.py
```diff
@@ -0,0 +1,17 @@
...
+import numpy as np
+from numpy.typing import NDArray
+
+ScalarType = Union[float, complex, int]
+IntNDArray = NDArray[np.integer]
+FloatNDArray = NDArray[np.floating]
+ComplexNDArray = NDArray[np.complexfloating]
+
+IntArrayVariant = Union[int | list[int] | IntNDArray]
+IntMatrixVariant = Union[IntArrayVariant | list[list[int]]]
```
ricott1: Is this used in the codebase? Same for some other types

ricott1 (commented):
In general, very much in favour of this. Probably we should also add some types form circuits and gates

.gitignore
f-martini: reverted 👍🏼

RickSimon79221 (approved):

ricott1 (approved):

PR #128, opened 2026-02-16
Fix pauli symmetries

RickSimon79221:
## 📝 Description
In previous big merge, Pauli symmetry code was not updated to new syntax. That is done here. Also fixes bug in the generation of random pauli-symmtric hamiltonians with redundant qubits. Also adds documentation and tests for pauli symmetry code 

- [x] The code passes CI.


src/sympleq/core/symmetries/pauli.py
```diff
@@ -1,15 +1,26 @@
 import numpy as np
 import itertools
 from sympleq.core.paulis import PauliSum
-from sympleq.core.circuits import Circuit, Hadamard as H, SUM as CX, PHASE as S
+from sympleq.core.circuits import Circuit, HADAMARD as H, CX, PHASE as S
```
ricott1: I think we should import `GATES` and use the gates as `GATES.H` rather than instatiating new singletons (i give an example later)

src/sympleq/core/symmetries/pauli.py
```diff
@@ -40,53 +85,133 @@ def number_of_S(x_exp, z_exp, d):
...
         if pauli_sum.x_exp[pauli_index, i]:
             number_of_sum_x = number_of_SUM_X(pauli_sum.x_exp[pauli_index, qudit],
                                               pauli_sum.x_exp[pauli_index, i],
                                               pauli_sum.dimensions[i])
-            list_of_gates += [CX(qudit, i, pauli_sum.dimensions[qudit])] * number_of_sum_x
-    C.add_gate(list_of_gates)
-    for g in list_of_gates:
-        pauli_sum = g.act(pauli_sum)
+            for _ in range(number_of_sum_x):
+                C.add_gate(CX(), qudit, i)
```
ricott1:
```suggestion
                C.add_gate(GATES.CX, qudit, i)
```

src/sympleq/core/symmetries/pauli.py
```diff
@@ -40,53 +85,133 @@ def number_of_S(x_exp, z_exp, d):
...
+                pauli_sum = CX().act(pauli_sum, (qudit, i))
```
ricott1:
```suggestion
                pauli_sum = GATES.CX.act(pauli_sum, (qudit, i))
```

src/sympleq/core/symmetries/pauli.py
```diff
@@ -40,53 +85,133 @@ def number_of_S(x_exp, z_exp, d):
...
+    C : Circuit
+        The circuit after the gates have been added to cancel out the Z part of the Pauli operator.
+    """
+    C.add_gate(H(), qudit)
+    pauli_sum = H().act(pauli_sum, qudit)
     for i in range(qudit + 1, q_max):
         if pauli_sum.z_exp[pauli_index, i]:
             number_of_sum_z = number_of_SUM_Z(pauli_sum.z_exp[pauli_index, i],
-                                              pauli_sum.x_exp[pauli_index, qudit],
+                                              pauli_sum.z_exp[pauli_index, qudit],
```
ricott1, line 152: this seems a substantial change, Im not familiar with this code, but are u sure we need to change x to z here?

src/sympleq/core/symmetries/pauli.py
```diff
@@ -40,53 +85,133 @@ def number_of_S(x_exp, z_exp, d):
...
+    number_of_phase = number_of_S(pauli_sum.x_exp[pauli_index, qudit], pauli_sum.z_exp[pauli_index, qudit],
+                                  pauli_sum.dimensions[qudit])
+    for _ in range(number_of_phase):
+        C.add_gate(S(), qudit)
+        pauli_sum = S().act(pauli_sum, qudit)
     return pauli_sum, C
 
 
-def cancel_pauli(P, current_qudit, pauli_index, circuit, n_q_max):
+def cancel_pauli(P: PauliSum, current_qudit: int, pauli_index: int, circuit: Circuit, n_q_max: int):
```
ricott1: missing return types

src/sympleq/core/symmetries/pauli.py
```diff
@@ -138,9 +303,8 @@ def symplectic_reduction_iter_qudit_(P, C, pivots, current_qudit):
     # does the current qudit have any X or Z components?
     if any(P.x_exp[:, current_qudit]) or any(P.z_exp[:, current_qudit]):
         if not any(P.x_exp[:, current_qudit]):  # If it is z we need to add a Hadamard gate to make it an X
-            g = H(current_qudit, P.dimensions[current_qudit])
-            C.add_gate(g)
-            P = g.act(P)
+            C.add_gate(H(), current_qudit)
```
ricott1: We could also add a new API as discussed (in a different PR), to do smtg like `C.H(current_qudit)`, what you think?

src/sympleq/core/symmetries/pauli.py
ldellant: Good catch, tbh I am not sure whether before was wrong or not. I think we never tested this function cause we are always diagonalizing in the Z basis. _If it is fast_ we could add a test where we diagonalize in the X basis and we check that all Z part is zero...

ldellant: Actually not sure about this, looking at `cancel_pauli`... Just ignore me :)

src/sympleq/core/symmetries/pauli.py
ldellant: Ignore this comment, but @ricott1 if you could explain why it is missing I would be happy :) 

Seems it returns a PauliSum and a Circuit, which are initialized?

src/sympleq/core/symmetries/pauli.py
```diff
@@ -152,31 +316,55 @@ def symplectic_reduction_iter_qudit_(P, C, pivots, current_qudit):
...
 
-        g = H(current_qudit, P.dimensions[current_qudit])
-        C.add_gate(g)
-        P = g.act(P)
+        C.add_gate(H(), current_qudit)
+        P = H().act(P, current_qudit)
     return C, pivots
 
 
 def symplectic_pauli_reduction(hamiltonian: PauliSum) -> Circuit:
```
ldellant: Can we eliminate this function? Seems redundant, as we can call `symplectic_reduction_qudit` and select C?

src/sympleq/core/symmetries/pauli.py
```diff
@@ -152,31 +316,55 @@ def symplectic_reduction_iter_qudit_(P, C, pivots, current_qudit):
...
+    Returns
+    -------
+    Circuit
+        The Circuit that implements the symplectic reduction algorithm.
+    """
     C, pivots = symplectic_reduction_qudit(hamiltonian)
     return C
 
 
 def pauli_reduce(hamiltonian: PauliSum) -> tuple[PauliSum, list[PauliSum], Circuit, list]:
```
ldellant, line 330: I almost wrote you to eliminate this function cause you have written the class for conditional Hamiltonian, that is just much better... I think however it is worth keeping for testing purposes :)

src/sympleq/models/random_hamiltonian.py
```diff
@@ -181,8 +180,9 @@ def random_pauli_symmetry_hamiltonian(n_qudits: int, n_paulis: int, n_redundant=
 
     P = PauliSum.from_string(pauli_strings, dimensions=[2] * n_qudits, weights=weights, phases=phases)
 
-    g = Gate.from_random(n_qudits, 2)
-    P = g.act(P, 2)
+    if shuffle:
+        g = Gate.from_random(n_qudits, 2)
```
ldellant, line 184: let's go a bit heavier with the gates... `10 * n_qudits ** 2`

tests/core_tests/symmetries_tests/symmetries_test.py
```diff
@@ -159,3 +160,22 @@ def test_random_arbitrary_symmetry(self):
                 assert np.all(H_s.weights == H_out.weights)
 
                 assert c.act(H, all_qudit_indices).to_standard_form() == H.to_standard_form()
+
+    def test_random_pauli_symmetry(self):
+        n_tests = 10
+        n_qudits = 10
```
ldellant: `n_ qudits = rng.integers(5, 25)`

tests/core_tests/symmetries_tests/symmetries_test.py
```diff
@@ -159,3 +160,22 @@ def test_random_arbitrary_symmetry(self):
...
+        n_paulis = 50
```
ldellant: `n_ paulis = rng.integers(5 * n_qudits, 10 * n_qudits ** 2)`

ldellant (approved):

src/sympleq/core/symmetries/pauli.py
ricott1: I meant adding the explicit return type as `-> type` at the end of the line

tests/core_tests/symmetries_tests/symmetries_test.py
ricott1: In general, I don't like using randomness in unit tests, and I think we should rather cover as much as possible and then have separate fuzzing tests. but no strong opinion

src/sympleq/core/symmetries/pauli.py
RickSimon79221: Yes, before, all the paulis where first put into a list and only executet after being collected. This way of doing it seemed to be harder/more annoying to do with the new inplementation of the gates and circuits, so instead I apply the gates now directly, as they are collected. Specifically here, the first gate is a Hadamard that takes the X in the qudit used to eliminate everything else and turns it into a Z. So, if we don't apply the Hadamard immediatly, the X-exponent is the important quantity, if we do, then it becomes the Z-exponent

src/sympleq/core/symmetries/pauli.py
RickSimon79221: Yeah, could be useful

src/sympleq/models/random_hamiltonian.py
RickSimon79221: Its not a random circuit, its a random gate of given size. It is already as random as it can get (probably).

ricott1 (approved):

PR #129, opened 2026-02-16
Dev issue#120

### LINKED ISSUE #120: Remove unused code from circuits/ folder
### Describe the enhancement

Remove all unused code/files from the core/circuits folder, like `random_symplectic.py`, `phase_correction.py`, `known_circuits.py`, ...

### Motivation

Clean up the code base and eliminate unused functions.

### Current behavior

_No response_

### Proposed behavior

_No response_

### Additional context

_No response_

ldellant:
## 📝 Description

- [x] The code passes CI.

closes #120


src/sympleq/applications/measurement/covariance_graph.py
```diff
@@ -87,7 +87,7 @@ def degree(self, a) -> int:
         #     a - (int) - vertex for which degree should be returned
         # Outputs:
         #     (int) - degree of vertex a
-        return np.count_nonzero(self.adj[a, :])
+        return int(np.count_nonzero(self.adj[a, :]))
```
ldellant, line 90: @ricott1 is there a more elegant way to ensure the return is `int` without explicitly using `int(...)`?

src/sympleq/core/paulis/pauli_sum.py
```diff
@@ -1213,7 +1213,7 @@ def ordered_eigenspectrum(self) -> tuple[np.ndarray, np.ndarray]:
         m = self.to_hilbert_space().toarray()
 
         # Get eigenvalues and eigenvectors
-        val, vec = np.linalg.eigh(m)
+        val, vec = np.linalg.eig(m)
```
ldellant: @ricott1 before you ask, I switched `eigh` -> `eig` because we have seen that the first has quite severe rounding errors...

scripts/examples/generate_basis.py
```diff
@@ -1,11 +1,11 @@
...
 import numpy as np
 sys.path.append("./")
-from sympleq.paulis import PauliString, PauliSum
-from sympleq.circuits import Circuit
-from sympleq.circuits.utils import solve_modular_linear
-from sympleq.circuits.known_circuits import to_x, to_ix
-from sympleq.hamiltonian import random_pauli_hamiltonian, pauli_reduce
+from sympleq.core.paulis import PauliString, PauliSum
+from sympleq.core.circuits import Circuit
+from sympleq.core import solve_modular_linear
```
ldellant, line 6: @c-nation I honestly do not fully recall what has happened to `solve_modular_linear`... I think I have eliminated cause it is not used in core and perhaps when testing was giving some error?

src/sympleq/models/toric_code.py
```diff
@@ -91,12 +91,11 @@ def build_plaquette_ops(self) -> list[list[int]]:
             for j in range(max_j):
                 edges = []
                 # Bottom horizontal edge of plaquette at (i,j)
-                if True:
```
ldellant, line 94: There was any reason for that `True`?

tests/core_tests/paulis_tests/pauli_utils_test.py
```diff
@@ -1,2 +1,171 @@
...
+            assert np.isclose(mean_1, mean_2), (
+                f"Expected Hamiltonian mean to match direct computation. "
+                f"Got {mean_1} vs {mean_2} for state {state} and PauliSum:\n{P}"
+            )
+
+    def test_covariance_matrix(self):
+        for _ in range(N_tests):
+
+            dimensions = choose_random_dimensions(25)
+            m_size = int(np.prod(dimensions))
```
ldellant, line 92: is there a reason we need to use `int` here?

ldellant (commented):
Ready to be reviewed :)

src/sympleq/core/paulis/pauli_sum.py
cerisola: I do not understand what you mean by severe rounding errors. If the `m` is Hermitian, I find it *very* hard to believe that `eig` performs better than `eigh`. If anything, this to me is a very big red flag that something is wrong somewhere. I would not push such a change without understanding what is happening.

Plus, in any case, such a non-trivial change should not be hidden in a pr about cleaning up old and redundant code.

src/sympleq/applications/measurement/covariance_graph.py
cerisola: I've seen in several places that there are explicit casts to `int`. Isn't the return type of `np.count_nonzero` an `np.int64`? If so, why the need to cast at all? Does it really need to be a python `int` and not a numpy `int64`? It seems to add unnecessary non-trivial allocations.

src/sympleq/core/paulis/pauli_sum.py
ldellant: Good point, I was wondering whether to not include this in this PR or not... I think I will remove it. Regarding the error, it was not when calling this function, but when testing with Rick. There, se saw discrepancies at the third-ish significant digits. Maybe we simply have to add an option that increases accuracy? 

Anyhow I will remove from this PR

cerisola: Did you try with a different LAPACK driver for `eigh`?
Explicitly making `m` Hermitian can also help if there are some rounding off errors when building it.

src/sympleq/applications/measurement/covariance_graph.py
ldellant: yes I am not sure what is the problem here... It seems to me that `int` and `np.int64` are treated as different things (which is absolutely reasonable), and we are currently not super consistent throughout the code. It should get better with @f-martini new PR I think

src/sympleq/core/paulis/pauli_sum.py
ldellant: Considering that I searched for "LAPACK" online, that should answer your question. As of now, there is also a step where we convert the sparse matrix into an array... Let me start a new PR where I tried to avoid that :)

src/sympleq/applications/measurement/covariance_graph.py
cerisola: They are different types and indeed represent very different objects. `np.int64` is a "standard" fixed width int, while python's native `int` is a much larger and complex object. In fact, if the creation and storage of many ints appears in some performance-critical part of the code, I would discourage the use of python `int`s.
But if it does not matter, then yes, for the type hinting I would use a generic type as what Francesco is introducing as you mention.

scripts/experiments/symmetries/notebooks/time_SWAP_symmetry_finder.ipynb
```diff
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
```
cerisola, line 5: This is comment is not just about this single line but generally the whole `time_SWAP_symmetry_finder.ipynb` notebook. Why was the notebook changed? It seems that the actual content did not change but the outputs were just cleared and then saved again.
Also, the corresponding commit message associated with that file change is not particularly helpful ...

ldellant: yes to be honest I do not know what happened here... I think at some point I pressed the "Clear Notebook" Task Runner and it ended up in the commit. I do admit it does not speak highly of my computational skills...

ldellant: Is there actual issues associated with that?

ricott1 (changes requested):
I would clean up `gate_decomposition_to_circuit` as well. Starting from removing unused functions there and then looking if functions in cicuits/utils can be removed as well

src/sympleq/core/paulis/pauli.py
```diff
@@ -109,80 +109,6 @@ def from_string(cls, pauli_str: str, dimension: int = DEFAULT_QUDIT_DIMENSION) -
 
         return P
 
-    @classmethod
```
ricott1, line 112: I would either remove Pauli completely or leave it for the moment

src/sympleq/core/paulis/utils.py
```diff
@@ -1,43 +1,7 @@
...
-            z = int(match.group(2))
-            p = int(match.group(3)) if match.group(3) is not None else 0
-            local_symplectics.append((x, z))
-            phases.append(p)
-
-    symplectic = np.array(local_symplectics).T
-    return symplectic.flatten(), sum(phases)
 
 
 def check_mappable_via_clifford(PauliSum: PauliSum,
```
ricott1, line 7: Please rename the variable to lowercase, this is very confusing 😅

PR #136, opened 2026-02-25
modified ordered eigenspectrum to take sparse matrices as input. Now …

ldellant:
…it has additional parameter k to select number of eigenvector and eigenstates to return

## 📝 Description

- [x] The code passes CI.


c-nation (approved):

RickSimon79221 (approved):

cerisola:
The pr most likely can be merged as-is. With that said, I would like to note that when using `eigsh` to find the smallest eigenvalues of a matrix, ideally you would like to use a shift-and-invert method, especially if you know a lower bound on the smallest eigenvalue (not sure if that is easy to know here).

PR #139, opened 2026-02-27
Added stabilizer functionalities

### LINKED ISSUE #157: Generalize ballistic approach to unitary circuits (inclusive of noise)
### Describe the bug

Currently, when we reconstruct the density matrix from a sum of stabilizers, we add all density matrices together with their own probabilities. This works only for the case of a Clifford noise channel. However, in the more general case we should, for each possible noise combination, obtain the density matrix from the sum of the state vectors from the unitary circuit.

Image

### Steps to reproduce

NA

### Expected behavior

Ballistic approach should be generalized and the method to obtain a density matrix from it as well. For the noiseless unitary, for instance, `stabilizer_to_hilbert_space` should return a state, not a density matrix, such that we can add the STATES together FIRST and THEN build the density matrix. 

It is possibly a good idea to have a class for stabilizer states, where the sum of stabilizer states can be stored and then we can implement the correct density matrix (broadly, we likely want to work with density matrices) directly.

### Actual behavior

The noise can be added only to clifford circuits

### Environment

_No response_

### Additional context

_No response_

ldellant:
## 📝 Description

closes #157

- [ ] The code passes CI.


.vscode/settings.json
```diff
@@ -53,5 +53,9 @@
   "jupyter.notebookFileRoot": "${workspaceFolder}",
   "task.autoDetect": "off",
   "python-envs.pythonProjects": [],
-  "editor.inlineSuggest.enabled": true
+  "editor.inlineSuggest.enabled": true,
```
f-martini, line 56: to add at the end of .vscode/extensions.json so that we can keep track of useful tools
```
"sonarsource.sonarlint-vscode"
```

f-martini (approved):

ldellant:
#closes #157

ldellant:
Closed as it is fixed by @Shreya93ban

PR #146, opened 2026-03-13
Add noise models and noisy circuit evolution

ricott1:
## 📝 Description
Add a new folder to core to handle noise in a unified way. Noise can be added to circuits by specifying a noise model. Noise models act on PauliSum by selecting a stochastic quantum trajectory. We also add methods to handle the unitary evolution in Hilbert space.

- [x] The code passes CI.


src/sympleq/core/circuits/circuits.py
```diff
@@ -275,9 +292,37 @@ def from_file(cls, file_path: str | Path) -> Circuit:
...
+        noise_model : NoiseModel
+            The noise model to apply after each gate.
+
+        Returns
+        -------
+        Circuit
+            This circuit instance (for method chaining).
+        """
+        self.noise_model = noise_model
+        return self
```
ldellant, line 345: Sure it is because I am missing something here or it is explained later... We also have spoken about that so this comment is mostly a check. 

the noise is added after each gate, but it is characteristic of the gate (or at least can be) and only acts on the qudits affected by the gate (or specified by the noise model), correct? In other words, if we have different gates we can have different noises, right?

src/sympleq/core/circuits/circuits.py
```diff
@@ -405,9 +450,26 @@ def act(self, pauli: PauliSum) -> PauliSum:
...
+
+        Returns
+        -------
+        PauliObject
+            The transformed Pauli object after all gates (and noise) are applied.
+        """
         for gate, qudits in zip(self._gates, self._qudit_indices):
             pauli = gate.act(pauli, qudits)
+            if self.noise_model is not None:
+                pauli = self.noise_model.apply_quantum_trajectory(pauli, qudits)
```
ldellant: ok so if we add noise we return a single trajectory (unless we do in the hilbert space). I think it is very solid, and could be a feature of sympleq.

ldellant: Also very sweet that we apply the quantum trajectory even without noise, that makes a lot of sense and makes generalizations (to e.g. noise or even non-Clifford [future work]) very easy to implement :)

src/sympleq/core/circuits/circuits.py
```diff
@@ -423,11 +485,87 @@ def act_iter(self, pauli: PauliSum) -> Generator[PauliSum, None, None]:
...
+            The input density matrix.
+
+        Returns
+        -------
+        HilbertOperator
+            The density matrix after all gates (and noise) are applied.
+        """
+        for gate, qudits in zip(self._gates, self._qudit_indices):
+            key = (gate, qudits)
+            if key not in self._unitary_cache:
```
ldellant, line 579: does it make so much sense to have a cache? In the past I wqas doing something with the cache and ended up taking too much memory. 

Can we add an if statement for the cache, something like if `np.prod(dims) > NUMBER` do not chache anything? Here number would be something like RAM / (n_qudits**2)

src/sympleq/core/circuits/gates.py
```diff
@@ -149,14 +157,17 @@ def solve_from_target(cls, input_tableau: TableauLike, target_tableau: TableauLi
 
     @property
     def name(self) -> str:
+        """str : Human-readable name of the gate."""
```
ldellant, line 160: bip bop bip

src/sympleq/core/circuits/gates.py
```diff
@@ -453,7 +517,7 @@ def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
                     out_k = (j + k) % d
                 out_idx = j * d + out_k  # |j, out_k⟩
                 U[out_idx, in_idx] = 1.0
```
ldellant, line 536: We should do this directly as a sparse matrix rather than an np.array and then convert.

is this PR a place to fix that? I think that we have the definition of a CX in AQUIRE that is already in sparse format..

ldellant: let me know if this will be fixed in this PR and I'll write how to do it :)

src/sympleq/core/circuits/gates.py
```diff
@@ -483,7 +547,7 @@ def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
                 in_idx = j * d + k   # |j,k⟩
                 out_idx = k * d + j  # |k,j⟩
                 U[out_idx, in_idx] = 1.0
-        return sp.csr_matrix(U)
+        return HilbertOperator(U)
```
ldellant, line 567: see previous comment

src/sympleq/core/circuits/gates.py
```diff
@@ -517,7 +581,7 @@ def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
             for k in range(d):
                 idx = j * d + k
                 U[idx, idx] = omega ** (j * k)
-        return sp.csr_matrix(U)
+        return HilbertOperator(U)
```
ldellant, line 601: see previous comment

src/sympleq/core/circuits/gates.py
```diff
@@ -526,6 +590,156 @@ def inverse(self) -> _CZ:
...
+        phase_vector = np.array([0, 0], dtype=int)
+
+        super().__init__("Id", symplectic, phase_vector)
+
+    def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
+        if dimension is None:
+            dimension = DEFAULT_QUDIT_DIMENSION
+        d = dimension
+        # X|j⟩ = |j+1 mod d⟩, X^{-1}|j⟩ = |j-1 mod d⟩
+        U = np.eye(d, dtype=complex)
```
ldellant, line 627: same comment as before

src/sympleq/core/circuits/gates.py
```diff
@@ -526,6 +590,156 @@ def inverse(self) -> _CZ:
...
+        if dimension is not None:
+            return np.array([0, -2 * x], dtype=int) % (2 * dimension)
+        return np.array([0, -2 * x], dtype=int)
+
+    def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
+        if dimension is None:
+            dimension = DEFAULT_QUDIT_DIMENSION
+        d = dimension
+        # X|j⟩ = |j+1 mod d⟩, X^{-1}|j⟩ = |j-1 mod d⟩
+        U = np.zeros((d, d), dtype=complex)
```
ldellant, line 670: same comment as above

src/sympleq/core/circuits/gates.py
```diff
@@ -526,6 +590,156 @@ def inverse(self) -> _CZ:
...
+            return np.array([2 * z, -2 * x], dtype=int) % (2 * dimension)
+        return np.array([2 * z, -2 * x], dtype=int)
+
+    def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
+        if dimension is None:
+            dimension = DEFAULT_QUDIT_DIMENSION
+        d = dimension
+        omega = np.exp(2j * np.pi / d)
+        # Y = X * Z: Y|j⟩ = ω^j |j+1 mod d⟩
+        U = np.zeros((d, d), dtype=complex)
```
ldellant, line 710: same as above

src/sympleq/core/circuits/gates.py
```diff
@@ -526,6 +590,156 @@ def inverse(self) -> _CZ:
...
+        super().__init__(name, symplectic)
+
+    def phase_vector(self, dimension: int | None = None) -> PhasesType:
+        # h = 2 * Ω @ [0, z] = 2 * [z, 0]
+        z = self._tableau[1]
+        if dimension is not None:
+            return np.array([2 * z, 0], dtype=int) % (2 * dimension)
+        return np.array([2 * z, 0], dtype=int)
+
+    def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
```
ldellant, line 745: same as above - I think this is the last one, you can assign an issue to me about this, that I can do after this PR is merged :)

src/sympleq/core/circuits/gates.py
```diff
@@ -558,6 +772,24 @@ def __init__(self):
         self._CZ = _CZ()
         # CZ is self-inverse, already handled in the class
 
+        self._Id = _Id()
+
+        # Pauli gates
+        self._X = _X(is_inverse=False)
+        self._X_inv = _X(is_inverse=True)
+        self._X._inverse = self._X_inv
```
ldellant, line 797: Stupid question... why both `_inv` and `_inverse`?

src/sympleq/core/circuits/utils.py
```diff
@@ -143,9 +143,9 @@ def _multi_index_to_linear(index: list[int] | np.ndarray, dims: list[int] | np.n
     return idx
 
 
-def embed_unitary(U_local: HilbertOperator,
-                  qudit_indices: list[int] | np.ndarray,
-                  total_dimensions: DimensionsType) -> HilbertOperator:
+def embed_unitary(U_local: sp.csr_matrix,
```
ldellant: `sp.csr_matrix` -> new Type chosen for hilbert object (forgot the name)?

src/sympleq/core/circuits/utils.py
```diff
@@ -143,9 +143,9 @@ def _multi_index_to_linear(index: list[int] | np.ndarray, dims: list[int] | np.n
...
+                  qudit_indices: tuple[int, ...] | list[int] | np.ndarray,
+                  total_dimensions: DimensionsType) -> sp.csr_matrix:
```
ldellant: same as above

src/sympleq/core/noise/noise_model.py
```diff
@@ -0,0 +1,361 @@
...
+Phys. Rev. A 108, 062604 (2023). DOI: 10.1103/PhysRevA.108.062604
+
+Overview
+--------
+A quantum noise channel transforms a Pauli string PS as:
+
+    PS_out = Σ_{m,n} λ_{mn} σ_m PS σ†_n
+
+where σ_m are Pauli basis operators and λ is the process matrix.
+
```
ldellant: Maybe only write the Kraus form, since it looks like we are going with that?

src/sympleq/core/noise/noise_model.py
```diff
@@ -0,0 +1,361 @@
...
+    λ_{mn} = Σ_i α_{im} α*_{in}
+
+The probability of each Kraus operator (used for quantum trajectory sampling) is:
+
+    p_i = Σ_j |α_{ij}|² = ||K_i||²
+
+We take two simplifying assumptions:
+**Uncorrelated noise**
+For uncorrelated noise on multiple qudits, the multi-qudit quantities are
+tensor products of single-qudit quantities.
```
ldellant: I do not think we are taking this assumption... We simply choose the noise to be uncorrelated when we decide that a 2-qubits gate acts with (say) dephasing on both qubit. But we can define a 2-qubit noise that is correlated, I think...

ldellant: it seems so to me based on what I understand below...

src/sympleq/core/noise/noise_model.py
```diff
@@ -0,0 +1,361 @@
...
+
+        # E.g.: single qudit noise on single qudit gate
+        cum_prob = 0
+        if n_qudits == kraus_n_qudits:
+            for kraus_gate, probability in zip(self.kraus_gates(), self.kraus_probabilities()):
+                unitary = embed_unitary(kraus_gate.local_unitary(dimension), qudits, dimensions)
+                output_rho = _apply_unitary(output_rho, unitary, probability)
+                cum_prob += probability
+
+            assert np.abs(cum_prob - 1.0) < 10**(-5), f"cum prob {cum_prob}"
```
ldellant: `10**(-10)` to make sure?

src/sympleq/core/noise/noise_model.py
```diff
@@ -0,0 +1,361 @@
...
+            for kraus_gates, probabilities in zip(all_gates_combinations, all_probabilities_combinations):
+                unitary = embed_unitary(kraus_gates[0].local_unitary(dimension), (qudits[0],), dimensions)
+                for kraus_gate, qudit in zip(kraus_gates[1:], qudits[1:]):
+                    unitary = embed_unitary(kraus_gate.local_unitary(dimension), (qudit,), dimensions) @ unitary
+
+                probability = float(np.prod(probabilities))
+                output_rho = _apply_unitary(output_rho, unitary, probability)
+                cum_prob += probability
+
+            assert np.abs(cum_prob - 1.0) < 10**(-5), f"cum prob {cum_prob}"
```
ldellant: `10**(-10)`

ldellant (commented):
The monster woke up so I have not looked at the tests (will do later on), but concerning the remainder I have seen it alllllll :)

Well well done!

src/sympleq/core/circuits/circuits.py
ricott1: at the moment, in short, no (noise only cares about number of qudits on which the gate acts). But it would be simple to add in the future.

Longer explanation: currently, after each gate we apply the noise model defined for the whole circuit. If we want to have gate-specific noise, we need to tell the circuit which noise to apply at each gate. we can relatively easily also make it qudit specific (different noise on different qudits even for the same gate).

src/sympleq/core/circuits/circuits.py
ricott1: second comment: what you mean? if no noise is define, we do not apply the quantum trajectory

src/sympleq/core/circuits/circuits.py
ricott1: We can have the flag. The tradeoff here is memory (size of unitary of the gate for each gate) vs calculating the unitary at each step. We could even do smtg in between, and only do the cache if a gate is called several times, but that would need a first pass to count the gates, a bit more complex. 

To be completely honest, I struggle to see how this can become a memory bottleneck unless we talk about really big circuits.

src/sympleq/core/circuits/gates.py
ricott1: Since this is not part of the PR, it's better to open an issue (or a PR directly) to address this, but we shouldnt do it here

src/sympleq/core/circuits/gates.py
ricott1: what u mean? I don't understand what u are referring to

src/sympleq/core/circuits/gates.py
ricott1: one is the inverse gate directly, the other is the reference to the inverse in the X gate (there is a 'dot')

src/sympleq/core/circuits/utils.py
ricott1: ah yeah, I messed this up

src/sympleq/core/circuits/circuits.py
ldellant: ok. My point was: important to keep in mind for a future feature, and I think it is the case. Kudos!

src/sympleq/core/circuits/circuits.py
ldellant: Ah! Well, what I meant is that if we have no noise, we can still use `apply_quantum_trajectory` to get the outcome. Since it is deterministic, it is equivalent to use `act`. 

What I meant is that no matter what we are doing, it look to me that `apply_quantum_trajectory` can be used in any circumstance:
- (no noise & no non-Clifford) -> equivalent to `act`
- (noise & no non-Clifford) -> what you have done here
- (no noise & non-Clifford) -> future generalization, but still based on `apply_quantum_trajectory`
- (noise & non-Clifford) -> future generalization, but still based on `apply_quantum_trajectory`

What I meant is that `apply_quantum_trajectory` can be a very clean building block to simulate circuits more in general. By doing that well and clean we will have a very powerful tool that nobody has :)

src/sympleq/core/circuits/circuits.py
ldellant: Well I guess that for `n` qubits we have order `n^2` possible 2 qubits gates... Hence my `RAM / (n_qudits**2)`. Circuits of depths bigger than `n^2` are fairly common... If it is fast and easy, I would put the flag, otherwise ignore this comment as it was a bit pedantic

tests/core_tests/noise_tests/noise_model_test.py
```diff
@@ -0,0 +1,143 @@
...
+            DephasingNoise(0.1)
+        ])
+
+        single_ops = single.kraus_probabilities()
+        composite_ops = composite.kraus_probabilities()
+
+        # Same operators, same weights (since averaging identical models)
+        assert len(single_ops) == len(composite_ops)
+        for s_op, c_op in zip(single_ops, composite_ops):
+            assert np.isclose(s_op, c_op)
```
ldellant, line 146: An important test to include is consistency between the hilbert outcome and trajectories outcome, with convergence within error bars for different number of trajectories....

ldellant: Can be included later on

ldellant (approved):

src/sympleq/core/circuits/circuits.py
ricott1: Added the flag

tests/core_tests/noise_tests/noise_model_test.py
ricott1: will be part of the RMB PR I think

src/sympleq/core/circuits/circuits.py
```diff
@@ -275,9 +304,54 @@ def from_file(cls, file_path: str | Path) -> Circuit:
...
             return cls.from_string(f.read())
 
+    def with_noise(self, noise_model: NoiseModel | list[NoiseModel | None]) -> Circuit:
+        """
+        Attach a noise model and return self for chaining.
+
+        Parameters
+        ----------
+        noise_model : NoiseModel | list[NoiseModel | None]
+            A single noise model applied after every gate, or a list of
```
ldellant, line 324: nice

src/sympleq/core/noise/noise_model.py
ldellant: SO here for the description I would say `ONE assumption` and would remove the assumption of uncorrelated noise... We can always define a composite gate and a noise model for that...

src/sympleq/core/noise/noise_model.py
```diff
@@ -0,0 +1,416 @@
...
+    p_i = Σ_j |α_{ij}|² = ||K_i||²
+
+We take two simplifying assumptions:
+**Uncorrelated noise**
+For uncorrelated noise on multiple qudits, the multi-qudit quantities are
+tensor products of single-qudit quantities.
+
+**Clifford noise**
+For now, we assume that Kraus operators are Clifford. This simplifies greatly how they act on PauliSums,
+as they are basically Gates.
```
ldellant: ... `as they are gates for each trajectori (see apply_quantum_trajectory)`

src/sympleq/core/noise/noise_model.py
```diff
@@ -0,0 +1,416 @@
...
+            return NotImplemented
+        return str(self) == str(other)
+
+    @classmethod
+    def from_string(cls, s: str) -> NoiseModel | None:
+        if s == "None":
+            return None
+        if s == "Noiseless":
+            return Noiseless()
+
```
ldellant, line 110: Is there a difference between `None` and ` "Noiseless" `? They may become the same thing no? To unify the framework, we could always assume that there is also noise, and no noise becomes noiseless.... I think it is a very minor thing tbh

src/sympleq/core/noise/noise_model.py
```diff
@@ -0,0 +1,416 @@
...
+        def _apply_unitary(output_rho: HilbertOperator | None, unitary: HilbertOperator,
+                           probability: float) -> HilbertOperator:
+            if output_rho is None:
+                output_rho = probability * (unitary @ rho @ unitary.conjugate().transpose())
+            else:
+                output_rho += probability * (unitary @ rho @ unitary.conjugate().transpose())
+            assert output_rho is not None
+            return output_rho
+
+        dimension = dimensions[qudits[0]]
```
ldellant: is this assuming all qudits have the same dimensions?

In general is not the case... BUT there exist not Clifford gates between qudits with different dimensions. So I thing that we can keep this assumption, at least for a long long time...

src/sympleq/core/noise/noise_model.py
```diff
@@ -0,0 +1,416 @@
...
+        n_qudits = len(qudits)
+
+        kraus_n_qudits = self.n_qudits()
+        if n_qudits < kraus_n_qudits:
```
ldellant, line 225: maybe add assert to yield error?

tests/core_tests/circuits_tests/circuits_test.py
```diff
@@ -439,6 +440,73 @@ def test_roundtrip_preserves_behavior(self):
 
         assert original_result == restored_result
 
+    def test_roundtrip_with_global_noise(self):
```
ldellant, line 443: looking at these tests, I just realized that equality is not super obvious... Are two circuits equal if they have the same gates, if they have the same tableau and phases, or if they have the same gates AND the same noise?

For the first two, we should already have a function that distinguish, and we agreed: equality means same circuit, then specific function that tells whether two circuits act identically or not. These two function should ignore noise, I think. Otherwise it becomes a bit messy...

Then, I would have another function, `equal_with_noise`, that yields true iff `circuit_1 = circuit_2` (gate-by-gate) and also all noises are exactly the same...

ldellant (approved):

src/sympleq/core/circuits/circuits.py
ricott1: can we close this comment?

src/sympleq/core/noise/noise_model.py
ricott1: I do assume that, maybe a comment would be useful. The reason is what u wrote, I thought Clifford gates on multiple qudits assumed homogeneous dimensions. In general, like for Pauli gates, they could have different dimensions tho, but then this is sort of an 'upstream' problem in the Gate class (no way of getting unitary for non-homogeneous dimensions)

tests/core_tests/circuits_tests/circuits_test.py
ricott1: Note, here we are testing the serialization/deserialization path, nothing regarding equalities of circuits per se

src/sympleq/core/noise/noise_model.py
ricott1: I would keep it flexible. I added the Noiseless class in case it was useful for some interfaces

src/sympleq/core/noise/noise_model.py
ldellant: I agree... I think that the best is to leave a comment now, and perhaps generalize in the future. 

Even better than a comment, actually, would be an `AssertionError` that triggers if qudits do not have the same dimensions... This way the user would know that they are doing something wrong...

src/sympleq/core/noise/noise_model.py
ldellant: ah ok I already saud that hehe ohoh

tests/core_tests/circuits_tests/circuits_test.py
ldellant: yes indeed let me open an issue for that

src/sympleq/core/noise/noise_model.py
```diff
@@ -0,0 +1,479 @@
...
+        probs = np.cumsum(self.kraus_probabilities())
+        idx = int(np.searchsorted(probs, self.rng.random()))
+        kraus_gate = self.kraus_gates()[idx]
+        if kraus_gate.n_qudits == n_qudits:
+            pauli_sum = kraus_gate.act(pauli_sum, qudits)
+        elif kraus_gate.n_qudits == 1:
+            # Act independently on each qudit
+            for qudit in qudits:
+                pauli_sum = kraus_gate.act(pauli_sum, qudit)
+        # FIXME: handle generic case, else:...
```
RickSimon79221, line 179: Is this FIXME important / has it been handled?

src/sympleq/core/circuits/circuits.py
```diff
@@ -55,6 +57,10 @@ def __init__(self, dimensions: DimensionsType,
         self._gates = gates
         self._qudit_indices = list(qudit_indices)
 
+        self._noise_model_per_gate: list[NoiseModel | None] = [None] * self.n_gates()
+        self._use_unitary_cache = use_unitary_cache
+        self._unitary_cache: dict[tuple, tuple[HilbertOperator, HilbertOperator]] = {}
```
RickSimon79221: Are all HilbertOperator-type variables proper inndeces for a dictionary?

RickSimon79221 (approved):

src/sympleq/core/circuits/circuits.py
```diff
@@ -374,7 +461,11 @@ def __add__(self, other: Circuit) -> Circuit:
 
         new_gates = self._gates + other._gates
         new_qudits = self._qudit_indices + other._qudit_indices
```
ldellant, line 463: `new_qudit` -> `new_qudit_indices` (finally I can be pedantic as well!)

src/sympleq/core/circuits/circuits.py
```diff
@@ -596,8 +784,8 @@ def to_string(self) -> str:
         '{"dimensions": [2, 3], "data": [["H", [0]], ["CX", [0, 1]]]}'
         """
         gate_data = []
-        for gate, qudits in zip(self._gates, self._qudit_indices):
-            gate_data.append([gate.name, [int(q) for q in qudits]])
+        for gate, qudits, noise in zip(self._gates, self._qudit_indices, self._noise_model_per_gate):
+            gate_data.append([gate.name, [int(q) for q in qudits], noise.__str__()])
         return json.dumps({"dimensions": [int(d) for d in self.dimensions], "data": gate_data})
```
ldellant, line 795: Stupid question... Isn't this redundant given that there is already `__str__` above? Ignore this comment if not :)

src/sympleq/core/circuits/gates.py
ldellant: badabop

src/sympleq/core/circuits/gates.py
```diff
@@ -144,14 +157,17 @@ def solve_from_target(cls, input_tableau: TableauLike, target_tableau: TableauLi
...
+        """str : Human-readable name of the gate."""
         return self._name
 
     @property
     def n_qudits(self) -> int:
+        """int : Number of qudits the gate acts on."""
         return self._n_qudits
 
     @property
     def symplectic(self) -> TableauType:
```
ldellant, line 169: just for me... For the `PauliObject` class we call this `tableau` correct? Maybe we can use the same name here? Or keep it different for clarity... good question, good question...

src/sympleq/core/circuits/circuits.py
```diff
@@ -425,11 +536,88 @@ def act_iter(self, pauli: PauliSum) -> Generator[PauliSum, None, None]:
...
+        Parameters
+        ----------
+        rho : HilbertOperator
+            The input density matrix.
+
+        Returns
+        -------
+        HilbertOperator
+            The density matrix after all gates (and noise) are applied.
+        """
```
ldellant, line 576: Wouldn't it be better to use `act_in_hilbert_space` as defined for the `Gate` class iteratively here, to use less functions? 

Is this different solution used because of `_unitary_cache`; if we use `act_in_hilbert_space` as defined for the `Gate` class we cannot store matrices?

Perhaps a solution is to modify `act_in_hilbert_space` as defined for the `Gate` by giving an extra boolean input parameter `return_unitary: bool = False` such that if true it also return the unitary? Or some solution on these lines?

src/sympleq/core/circuits/gates.py
ldellant: If I recall correctly... YES I think I do :) 

Essentially, By doing this way we build an array that then is transformed into a sparse matrix. For efficiency, it would be much better to build the sparse matrix from the beginning. This would also eliminate the `for: for:` loop since the indices could be generated in an array directly. 

The way of doing so is in the Pauli operator matrices, these are generated already sparse. I can do that after this PR is merged, you can assign a ticket to me :)

src/sympleq/core/circuits/gates.py
ldellant: indeed, indeed 

Sorry you will be missing our movie night :( 
Have you seen "Smetto quando voglio"? Can you provide a feedback to the others saying it is an incredibly appropriate movie?

src/sympleq/core/circuits/gates.py
ldellant: it is the same comment... recall about the movie suggestion...

src/sympleq/core/circuits/gates.py
ldellant: and again - oh my oh my

src/sympleq/core/circuits/gates.py
ldellant: kill me

src/sympleq/core/circuits/gates.py
ldellant: ah whoah I already said that... solid solid

src/sympleq/core/circuits/gates.py
ldellant: totally unclear but you have my total trust

ldellant (approved):

src/sympleq/core/circuits/gates.py
ricott1: This would be unrelated to this PR, better to do it in a separate one

src/sympleq/core/circuits/circuits.py
ricott1: the key needs to be hashable, but here the HilbertOperator is the value. I improved the typing to clarify the key type

src/sympleq/core/circuits/gates.py
ricott1: we can do that, but it's unrelated to this PR

src/sympleq/core/circuits/circuits.py
ricott1: This is more a string encoding, so that we can later decode it

src/sympleq/core/noise/noise_model.py
ricott1: I improved the comment. I don't think this is needed for now, it handles cases like a 2-qudit noise on a multiple (>2) qudits gate (like a circuit compound gate)

src/sympleq/core/circuits/circuits.py
ricott1: unrelated

src/sympleq/core/noise/noise_model.py
ricott1: done

src/sympleq/core/circuits/circuits.py
ricott1: good point. I added an extensive FIXME for this since it does require some thinking about using the cache and delegating to gate method (we need to do this consistently also for the noise, which at the moment does no have a cache)

PR #147, opened 2026-03-15
Add `BayesianEstimator` class

ricott1:
## 📝 Description
Add a Bayesian estimation utility class, which manages running stochastic jobs until a certain maximum variance has been achieved.

- [x] The code passes CI.

### DIFF
--- a/src/sympleq/core/bayesian_estimation.py
+++ b/src/sympleq/core/bayesian_estimation.py
@@ -1,5 +1,4 @@
-import math
-from typing import Callable, Hashable
+from typing import Any, Callable, Generator, Hashable
 
 
 class BayesianEstimator:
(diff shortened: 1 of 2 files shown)


ldellant (approved):
Long term would be nice to integrate with AQUIRE, but atm let's keep them separeted :)

src/sympleq/core/bayesian_estimation.py
```diff
@@ -0,0 +1,72 @@
+from typing import Any, Callable, Generator, Hashable
+
+
+class BayesianEstimator:
+    def __init__(self, threshold: float = 10**(-2), min_runs: int = 100) -> None:
```
RickSimon79221: Comment after DSU to add a max_run cutoff (basically a bound at which the estimation is stopped if it grows too big)

RickSimon79221 (approved):

src/sympleq/core/bayesian_estimation.py
ricott1: done

PR #155, opened 2026-04-01
overloads on Gate.act

c-nation:
## 📝 Description

Tiny change. Just adds overloads so that gate.act is type preserving.

- [ x] The code follows the defined style guide.
- [ ] The code passes CI.


src/sympleq/core/circuits/gates.py
```diff
@@ -182,7 +183,19 @@ def __eq__(self, other):
...
+
+    @overload
+    def act(self, pauli: PauliString, qudits: int | tuple[int, ...]) -> PauliString:
+        ...
+
+    @overload
     def act(self, pauli: PauliObject, qudits: int | tuple[int, ...]) -> PauliObject:
+        ...
+
+    def act(self, pauli: PauliObject, qudits: int | tuple[int, ...]):
```
ricott1: I understand it's not relevant since we have the overload, but for clarity I would add the return type `-> PauliObject`. Also, OCD comment, but I would use the same order as in Circuit.act, forst PauliString and then PauliSum.

src/sympleq/core/circuits/gates.py
```diff
@@ -182,7 +183,19 @@ def __eq__(self, other):
...
 
+    @overload
+    def act(self, pauli: PauliSum, qudits: int | tuple[int, ...]) -> PauliSum:
+        ...
+
+    @overload
+    def act(self, pauli: PauliString, qudits: int | tuple[int, ...]) -> PauliString:
+        ...
+
+    @overload
```
ldellant, line 221: Not very skilled in using `@overload` at all... But the type of `PauliObject` is't inclusive of `PauliString` and `PauliSum` as well? In other words, do we need to specify these differences? I thought that the idea of the `PauliObject` class was to not needing having to specify all cases...

```
from typing import TypeVar
PauliType = TypeVar("PauliType", bound=PauliObject)

def act(self, pauli: PauliType, qudits: int | tuple[int, ...]) -> PauliType:
```

src/sympleq/core/paulis/pauli_object.py
```diff
@@ -776,7 +776,7 @@ def __hash__(self) -> int:
              tuple(self.dimensions.tobytes()))
         )
 
-    def __dict__(self) -> dict:
+    def get_dict(self) -> dict:
```
ricott1, line 779: why did we do this change? seems to have unintended effects no?

src/sympleq/core/circuits/gates.py
```diff
@@ -262,8 +265,10 @@ def act(self, pauli: PauliObject, qudits: int | tuple[int, ...]) -> PauliObject:
 
         new_phases = (pauli.phases + acquired_phases) % (2 * pauli.lcm)
 
-        return pauli.__class__(tableau=new_tableau, dimensions=pauli.dimensions,
-                               weights=pauli.weights, phases=new_phases)
+        return cast(PauliType, pauli.__class__(
```
ricott1: I'm sorry if im going back and forth on this relatively non-consequential change, but what u had before seemed better to me, as it was more inline with the circuit class, smtg like

```python
    @overload
    def act(self, pauli: PauliSum, qudits: int | tuple[int, ...]) -> PauliSum:
        ...

    @overload
    def act(self, pauli: PauliString, qudits: int | tuple[int, ...]) -> PauliString:
        ...

    @overload
    def act(self, pauli: PauliObject, qudits: int | tuple[int, ...]) -> PauliObject:
        ...

    def act(self, pauli, qudits):
        <implementation here>
```

src/sympleq/core/circuits/gates.py
```diff
@@ -974,7 +979,7 @@ def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
         z = self.pauli_string.z_exp
         return pauli_unitary_from_tableau(d, x, z, convention="bare")
 
-    def act(self, pauli: PauliObject, qudits: int | tuple[int, ...] | None = None) -> PauliObject:
+    def act(self, pauli: PauliType, qudits: int | tuple[int, ...] | None = None) -> PauliType:
```
ricott1: samehere

src/sympleq/core/paulis/pauli_sum.py
```diff
@@ -1178,12 +1178,16 @@ def ordered_eigenspectrum(self, num_eigens: int | None = None) -> tuple[np.ndarr
 
         # Convert PauliSum to matrix form
         sparse_matrix = self.to_hilbert_space()
+        matrix_shape = sparse_matrix.shape
+        if matrix_shape is None:
```
ricott1, line 1182: when can this be the case?

c-nation: It shouldn't ever be the case, but the next line gives a warning if you remove it as scipy allows None as an output

src/sympleq/core/paulis/pauli_object.py
c-nation: __dict__ cannot be redefined with the ABC base class. It gives a warning.

src/sympleq/core/circuits/gates.py
c-nation: I agree overload is more readable. TypeVar is operationally the same and shorter, and that seemed like the consensus in the meeting. I'd be happy to go back to overloads.

ricott1: I think what I meant is to have the typevar in the implementation and keep the overload, sorry that was not so clear

src/sympleq/core/paulis/pauli_object.py
ricott1: I don't get the warnging to be honest, but fair enough (I dont think we ever use it til now)

src/sympleq/core/paulis/pauli_sum.py
ricott1: sounds good

ricott1 (approved):

f-martini (approved):

PR #163, opened 2026-04-25
Add constructor for circuit with fixed depth

ricott1:
## 📝 Description
Add a constructor to create circuits with fixed depth.

- [x] The code passes CI.

### DIFF
--- a/src/sympleq/core/circuits/circuits.py
+++ b/src/sympleq/core/circuits/circuits.py
@@ -302,7 +302,7 @@ class Circuit:
         n_qudits = len(dimensions)
         num_gates = n_qudits * depth
         # Divide by two since each gate applies to 2 qudits
-        num_two_qudits_gates = int(two_qudit_gate_ratio * num_gates) // 2
+        num_two_qudits_gates = int(two_qudit_gate_ratio * num_gates) // 2 if n_qudits > 1 else 0
 
         # First assign only 1-qudit gates
         layers: list[dict[tuple[int, ...], Gate]] = []
@@ -314,8 +314,7 @@ class Circuit:
                 layer[(q,)] = gate
             layers.append(layer)
 
-        # Distribute num_two_qudits_gates over depth layers randomly,
-        # with max max_two_qudits_gates_per_layer per layer.
+        # Distribute num_two_qudits_gates over depth layers randomly.
         if num_two_qudits_gates > 0 and two_qudit_gates:
             # available_qudits is a list of set_idxs (one per layer). Each element is a list of qudit indices,
             # indicating which qudits can be combined in a 2-qudit gate.
--- a/src/sympleq/core/circuits/circuits.py
+++ b/src/sympleq/core/circuits/circuits.py
@@ -298,7 +298,6 @@ class Circuit:
         for i, val in enumerate(dimensions):
             groups[val].append(i)
         two_qudits_gates_index_sets = [v for v in groups.values() if len(v) >= 2]
-        num_index_sets = len(two_qudits_gates_index_sets)
 
         n_qudits = len(dimensions)
         num_gates = n_qudits * depth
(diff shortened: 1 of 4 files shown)


src/sympleq/core/circuits/circuits.py
```diff
@@ -244,6 +244,117 @@ def from_gates_and_qudits(cls, dimensions: DimensionsLike,
...
+
+        Returns
+        -------
+        Circuit
+            A new random Circuit.
+        """
+
+        if gates_set is None:
+            single_qudit_gates: list[Gate] = [GATES.H, GATES.S]
+            two_qudit_gates: list[Gate] = [GATES.CX, GATES.CZ, GATES.SWAP]
```
ldellant, line 279: I would keep it simple and only have CZ or CX here.

src/sympleq/core/circuits/circuits.py
```diff
@@ -244,6 +244,117 @@ def from_gates_and_qudits(cls, dimensions: DimensionsLike,
...
+
+        if two_qudit_gate_ratio < 0.0 or two_qudit_gate_ratio > 1.0:
+            raise ValueError(
+                f"Invalid two_qudit_gate_ratio, it should be between 0 and 1 (got {two_qudit_gate_ratio}).")
+
+        if rng is None:
+            rng = default_rng()
+
+        dimensions = np.asarray(dimensions, dtype=int)
+        # generate list of lists of indexes for each dimension
```
ldellant, line 296: add a `raise ValueError` if there is only one qudit

ldellant: I would define n_qudits (line 302) here as well...

src/sympleq/core/circuits/circuits.py
```diff
@@ -244,6 +244,117 @@ def from_gates_and_qudits(cls, dimensions: DimensionsLike,
...
+        for _ in range(depth):
+            layer = {}
+
+            for q in range(n_qudits):
+                gate = single_qudit_gates[rng.integers(0, len(single_qudit_gates))]
+                layer[(q,)] = gate
+            layers.append(layer)
+
+        # Distribute num_two_qudits_gates over depth layers randomly,
+        # with max max_two_qudits_gates_per_layer per layer.
```
ldellant: I do not see a parameter called `max_two_qudits_gates`... If there is ignore this comment. It seems not necessary but if it is useful absolutely keep it :)

tests/core_tests/circuits_tests/circuits_test.py
```diff
@@ -603,6 +603,87 @@ def test_from_random_gates_set_invalid_type(self):
         with pytest.raises(ValueError):
             _ = Circuit.from_random(5, [2, 2], gates_set=GATES.H)  # type: ignore[arg-type]
 
+    def test_from_depth_default_gate_set(self):
+        """Default from_depth gate set is {H, S, CX, CZ, SWAP}."""
```
ldellant, line 607: Now H, S, CZ

ldellant (approved):
solid. Maybe a future extension is to insert a parameter called "density", where we include also the identity gate,.. but it is not relevant for now...

src/sympleq/core/circuits/circuits.py
ricott1: I dont think we need to raise an error, but it's true the check can skip the 2 qudit gates completely

src/sympleq/core/circuits/circuits.py
ricott1: good catch, it was an old comment

tests/core_tests/circuits_tests/circuits_test.py
ricott1: Not sure I follow here

RickSimon79221 (approved):

PR #177, opened 2026-06-24
Change `Setup Dev Environment` to include optional installations

ldellant:
## 📝 Description
With the RBM application we are importing packages that are not necessary for sympleq core functions and that have conflicts with the packages onto which sympleq relies. 

Now, when `Setup Dev Environment` task is run, it ask whether to install the optional dependencies and uses `uv` to override dependencies that would otherwise create conflicts. 

- [x] The code passes CI.


ldellant:
@f-martini would be great if you could look at this. I used AI to change the `Setup Dev Environment` task, so it would be great if someone that actually understands it could look at it. Few thoughts:

1) I have modified the Mac/Linux task as well, but since I could not try it out I do not know whether it worked
2) For overriding the dependencies I am relying on a package called `uv`. At the moment, `Setup Dev Environment` rely on `pip` (as before) for installing some packages and on `uv` for others. Would it be better to only use `uv`?
3) I have also modified the pipeline for the `venv` environment (it *should* now check whether there is a venv environment already, if not create one, use it and install the dependencies). I am not sure my changes work for all machines or only for mine...

.vscode/settings.json
```diff
@@ -43,6 +43,10 @@
   "esbonio.sphinx.confDir": "${workspaceFolder}/docs",
   "esbonio.sphinx.buildDir": "${workspaceFolder}/docs/_build",
 
+  // Python interpreter and terminal activation
+  "python.defaultInterpreterPath": "${workspaceFolder}/venv/Scripts/python.exe",
```
f-martini: This will not work on Linux and macOS, this line sould be substituted with:
```
"python.defaultInterpreterPath": "${workspaceFolder}/venv",
```
if that doesnt work try with the more verbose options:
```
    "windows": {
        "python.defaultInterpreterPath": "${workspaceFolder}/venv/Scripts/python.exe"
    },

    "linux": {
        "python.defaultInterpreterPath": "${workspaceFolder}/venv/bin/python"
    },

    "osx": {
        "python.defaultInterpreterPath": "${workspaceFolder}/venv/bin/python"
    }
```

scripts/tools/shell/setup_environment.bat
```diff
@@ -7,17 +7,35 @@ call env.bat
...
 
 REM Initializing virtual environment...
-if not exist %SRC_VENV% (
+if not exist %SRC_VENV%\Scripts\activate.bat (
     echo Creating virtual environment %SRC_VENV%...
-    python -m venv %SRC_VENV%
-    call %SRC_VENV%/Scripts/activate
-    call python -m pip install -r %DEV_REQUIREMENTS%
-    call deactivate
+    if exist %SRC_VENV% (rmdir /s /q %SRC_VENV% 2>nul)
```
f-martini: move this up, outside of the if statement,  if you always want to rebuild the dev environment from scratch (I agree that this might be the best option when running this script). Then there is no need to check for the existence of the activate .bat file. In the end it should be something like:
```
REM Initializing virtual environment...
if exist %SRC_VENV% (rmdir /s /q %SRC_VENV% 2>nul)
echo Creating virtual environment %SRC_VENV%...
python -m venv %SRC_VENV%
```

scripts/tools/shell/setup_environment.bat
```diff
@@ -7,17 +7,35 @@ call env.bat
...
+    powershell -NonInteractive -Command "py -m venv '%SRC_VENV%'"
+    if not exist %SRC_VENV%\Scripts\activate.bat (
+        echo ERROR: Failed to create virtual environment.
+        echo Ensure Python is installed, then run this task again.
+        exit /b 1
+    )
 )
 
-call %SRC_VENV%/Scripts/activate
+call %SRC_VENV%\Scripts\activate.bat
```
f-martini: Restore the original

scripts/tools/shell/setup_environment.sh
```diff
@@ -11,14 +11,26 @@ cd "$PROJECT_ROOT"
...
     echo "Creating virtual environment $SRC_VENV..."
     python -m venv "$SRC_VENV"
-    source "$SRC_VENV/bin/activate"
-    python -m pip install -r "$DEV_REQUIREMENTS"
-    deactivate
 fi
 
 source "$SRC_VENV/bin/activate"
 python -m pip install --upgrade pip setuptools setuptools-scm
+python -m pip install -r "$DEV_REQUIREMENTS"
```
f-martini: move dev requirments in [project.optional-dependencies] as done for the other optional dependencies.

.vscode/tasks.json
```diff
@@ -23,7 +23,13 @@
             "label": "Setup Dev Environment",
             "type": "shell",
             "windows": {
-                "command": "${workspaceFolder}/scripts/tools/shell/setup_environment.bat",
+                "command": "if (-not (Test-Path 'venv\\Scripts\\activate.bat')) { if (Test-Path 'venv') { Remove-Item -Recurse -Force venv }; py -m venv venv }; & '${workspaceFolder}/scripts/tools/shell/setup_environment.bat'",
```
f-martini: Restore. the original. The logic can be (and should be) added inside the script directly

scripts/configs/dev_requirements.txt
```diff
@@ -1,13 +1,11 @@
+uv>=0.5.0
```
f-martini: Move all the remaining dev dependencies into the `pyproject.toml`
Use something like what you already did for quantinuum, adding a new group:
```
[project.optional-dependencies]
development =  = [ 
    ...
]
```
It's finally time to say goodbye to the dev_requirements.txt :)

scripts/tools/shell/setup_environment.bat
```diff
@@ -7,17 +7,35 @@ call env.bat
...
 call python -m pip install --upgrade pip setuptools setuptools-scm
+call python -m pip install -r %DEV_REQUIREMENTS%
```
f-martini: Let's try to remove the dev_requirements.txt file and put the remaining requirements in the pyproject.toml. Then install them as a first step:

call python -m pip install -e ".[development]"

pyproject.toml
```diff
@@ -21,7 +21,15 @@ license = { text = "MIT" }
 requires-python = ">=3.11"
 
 [project.optional-dependencies]
-quantinuum = ["pytket>=1.0.0", "pytket-quantinuum>=0.58.0", "qnexus>=0.43.0"]
+quantinuum = ["pytket>=2.16.0", "pytket-quantinuum>=0.58.0", "qnexus>=0.43.0"]
```
f-martini: Love this, let's add dev requirements as a `development` optional dependency group :)

f-martini (commented):
I am fine with the idea behind these changes, but I would at least try the more “naive” approach of simply adding the Quantinuum dependency to Sympleq. In particular, I would expect that by specifying botorch>=0.18.0 in the Sympleq requirements (along with everything else), we would resolve our issues.

In particular, given that we currently have Quantinuum imports in our Sympleq source folder, it qualifies 100% as a dependency. It is not clear to me why it has been marked as an optional dependency. We will take care of removing it from the dependencies if we eventually split the application folder from the Sympleq base library. At the moment, we need it, and it is not optional at all.

.vscode/settings.json
ldellant: I added the latter as the first was having few problems

scripts/tools/shell/setup_environment.bat
ldellant: It gives me an error message now that does not create any problem(?) 
I write this comment just to remind myself to check later...

ldellant: I have removed 
`python -m venv %SRC_VENV%`
as it is seemingly redundant with the following line
`powershell -NonInteractive -Command "py -m venv '%SRC_VENV%'"`

f-martini (approved):

ldellant:
@f-martini regarding your last comment, I am not fully sure I understand it. The original issue ( I think) was that botorch>=0.18.0 is in conflict with aepsych, so it only appears when installing the optional dependencies for the RBM application. 

The other point, namely that we may want quantinuum packages not to be optional, may need a bit more care. While you are totally right in saying that what is in core is not optional, I also think that we want to keep sympleq as independent as possible from other packages. This goes towards the direction of better separating applications and core function. More about this today at the DSU perhaps :)

ldellant:
Using Python 3.14.3 environment at: venv
  × No solution found when resolving dependencies:
  ╰─▶ Because only aepsych<=0.8.0 is available and aepsych==0.8.0 depends on botorch==0.12.0, we can conclude that aepsych>=0.8.0 depends on
      botorch==0.12.0.
      And because sympleq[rbmalgorithms]==0.0.0 depends on aepsych>=0.8.0 and botorch>=0.18.0, we can conclude that sympleq[rbmalgorithms]==0.0.0 cannot
      be used.
      And because only sympleq[rbmalgorithms]==0.0.0 is available and you require sympleq[rbmalgorithms], we can conclude that your requirements are
      unsatisfiable.

RickSimon79221 (approved):

PR #178, opened 2026-06-29
Update core and Quantinuum integration from applications/rmb

ricott1:
## 📝 Description
Import core changes from RMB application

- [x] The code passes CI.


src/sympleq/core/circuits/circuits.py
```diff
@@ -244,6 +245,210 @@ def from_gates_and_qudits(cls, dimensions: DimensionsLike,
...
+
+        Returns
+        -------
+        Circuit
+            A new random Circuit.
+        """
+
+        if gates_set is None:
+            single_qudit_gates: list[Gate] = [GATES.H, GATES.S]
+            two_qudit_gates: list[Gate] = [GATES.CX, GATES.CZ, GATES.SWAP]
```
ldellant, line 279: Just use either the `CX` or the `CZ` gate, and eliminate the other two. That is a basis for Clifford and would keep everything tighter

src/sympleq/core/circuits/circuits.py
```diff
@@ -244,6 +245,210 @@ def from_gates_and_qudits(cls, dimensions: DimensionsLike,
...
+            gate = two_qudit_gates[rng.integers(0, len(two_qudit_gates))]
+            q_idxs = tuple(int(idx) for idx in rng.choice(range(n_qudits), 2, replace=False))
+            _gates.append(gate)
+            _qudit_indices.append(q_idxs)
+
+        shuffling_indices = list(range(len(_gates)))
+        rng.shuffle(shuffling_indices)
+
+        gates = [_gates[idx] for idx in shuffling_indices]
+        qudit_indices = [_qudit_indices[idx] for idx in shuffling_indices]
```
ldellant, line 334: smart solution :)

src/sympleq/core/circuits/circuits.py
```diff
@@ -244,6 +245,210 @@ def from_gates_and_qudits(cls, dimensions: DimensionsLike,
...
+
+        Returns
+        -------
+        Circuit
+            A new random Circuit.
+        """
+
+        if gates_set is None:
+            single_qudit_gates: list[Gate] = [GATES.H, GATES.S]
+            two_qudit_gates: list[Gate] = [GATES.CX, GATES.CZ, GATES.SWAP]
```
ldellant, line 373: Similar to above, let us only use `CZ` as entangling gate

src/sympleq/core/circuits/circuits.py
```diff
@@ -244,6 +245,210 @@ def from_gates_and_qudits(cls, dimensions: DimensionsLike,
...
+        # generate list of lists of indexes for each dimension
+        groups: dict[int, list[int]] = defaultdict(list)
+        for i, val in enumerate(dimensions):
+            groups[val].append(i)
+        two_qudits_gates_index_sets = [v for v in groups.values() if len(v) >= 2]
+
+        n_qudits = len(dimensions)
+        num_gates = n_qudits * depth
+        # Divide by two since each gate applies to 2 qudits
+        num_two_qudits_gates = int(two_qudit_gate_ratio * num_gates) // 2
```
ldellant: It is certainly a choice, but nobody in the field consider an entangling gate as 2 gates... 
Remove the `//2` at the end

src/sympleq/core/circuits/circuits.py
```diff
@@ -447,6 +711,18 @@ def n_gates(self) -> int:
...
+        Returns the number of 1-qubit gates in the circuit.
+        """
+        return len([g for g in self.gates if g.n_qudits == 1])
+
+    def n_2qb_gates(self) -> int:
+        """
+        Returns the number of 2-qubit gates in the circuit.
+        """
+        return len([g for g in self.gates if g.n_qudits == 2])
+
```
ldellant: minor detail: let us maintain a notation that is consistent throughout. Since we can use qudits let us never use qubits and always qudit. Therefore, `qb` -> `qd` in these functions' names...

src/sympleq/core/circuits/gates.py
```diff
@@ -661,6 +661,51 @@ def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
         return HilbertOperator(U)
 
 
+class _V(Gate):
+    """V = √X gate: X -> X, Z -> -Y = -XZ. Has special phase vector for qubits.
+
+    V is the X-axis analog of S: V = exp(-iπ/4 X). Together with S and any
+    entangling Clifford, V generates the single-qubit Clifford group, which
```
ldellant: not "single-qubit" Clifford group... It is the "qubit Clifford group"

ldellant: actually not even "qubit", since these definition work for any dimension (well done)... It is just the "Clifford group".

src/sympleq/core/circuits/gates.py
```diff
@@ -661,6 +661,51 @@ def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
...
+            name = "V"
+
+        super().__init__(name, symplectic, exceptional_phase_vectors=exceptional)
+
+    def local_unitary(self, dimension: int | None = None) -> HilbertOperator:
+        if dimension is None:
+            dimension = DEFAULT_QUDIT_DIMENSION
+        if dimension != 2:
+            raise NotImplementedError(
+                "V (= √X) is only implemented for qubits (dimension=2)."
```
ldellant, line 741: An easy way to define it for qudit is to apply a Hadamard to the S gate:

`H@S@H_inv`

(up to phase I think... it is not essential to implement this now, if you want we can do it or I can do it myself :) )

src/sympleq/core/noise/noise_model.py
```diff
@@ -475,6 +546,31 @@ def from_probabilities_and_gates(cls, probabilities: list[float], gates: list[Ga
             rng = default_rng()
         return cls(probabilities, gates, rng)
 
+    @classmethod
+    def from_paulis(cls, probabilities: list[float], rng: RNGGenerator | None = None) -> GenericNoise:
+        """
+        Create a GenericNoise from probabilities using the Pauli gates as default gates set.
```
ldellant, line 552: "... gates set. `probabilities[i]` refer to Pauli X, Y, and Z for `i`=1, 2, and 3, respectively"

src/sympleq/core/bayesian_estimation.py
```diff
@@ -69,6 +79,8 @@ def probability(self, key: Hashable) -> float:
         float
             Bayesian posterior mean probability.
         """
+        if key not in self._probabilities:
+            return 0.0
```
ldellant, line 85: If it is not in `self._probabilities` we want it to yield an error, not 0... We should not assign probabilities to quantities that are not being estimated. We risk providing a wrong estimate, as a zero probability of (say) getting +1 when measuring a Pauli has a very precise meaning.

src/sympleq/core/bayesian_estimation.py
```diff
@@ -85,6 +97,8 @@ def variance(self, key: Hashable) -> float:
         float
             Bayesian posterior variance.
         """
+        if key not in self._probabilities:
+            return 0.0
```
ldellant, line 103: Similar to previous comment, if it is not in `self._probabilities` it should yield an error.

src/sympleq/core/bayesian_estimation.py
ldellant: If I am misunderstanding something let's have a call

src/sympleq/core/bayesian_estimation.py
```diff
@@ -96,18 +110,77 @@ def _update_estimates(self, base: float, counts_tot: int, a_tot: float):
...
+        Returns
+        -------
+        dict[Hashable, int]
+            ``{outcome: count}`` for every observed outcome.
+        """
+        return dict(self._results)
+
+    def posterior_alpha_beta(self) -> tuple[float, float]:
+        """Return the Beta(1, 1) posterior parameters of the ``True`` outcome probability."""
+        return self._results.get(True, 0) + 1.0, self._results.get(False, 0) + 1.0
```
ldellant, line 157: Now here, if we have added `priors` to the class, instead of `+ 1.0` in the output we can `+ priors[0]` and `+priors[1]`, respectively.

src/sympleq/core/bayesian_estimation.py
```diff
@@ -1,5 +1,11 @@
...
 from typing import Any, Callable, Generator, Hashable
 
+from scipy.stats import beta as beta_dist
+
+
+type EstimatorCallable[T: Hashable] = Callable[[], T | list[T]]
+
 
 class BayesianEstimator:
     def __init__(self, threshold: float = 10**(-2), min_runs: int = 100, max_runs: int | None = None) -> None:
```
ldellant, line 13: add input `priors: [float, float] = [1.0, 1.0]` to make this more general without adding (much) complexity... Having the priors is generally important and we will likely change them at some point...

src/sympleq/core/bayesian_estimation.py
ldellant, line 39: self.priors = priors

src/sympleq/core/bayesian_estimation.py
```diff
@@ -33,6 +39,10 @@ def __str__(self) -> str:
         return (f"BayesianEstimator(threshold={self.threshold}, "
                 f"min_runs={self.min_runs}, max_runs={self.max_runs})")
```
ldellant, line 42: "..., priors ={self.priors})"

src/sympleq/core/bayesian_estimation.py
```diff
@@ -96,18 +110,77 @@ def _update_estimates(self, base: float, counts_tot: int, a_tot: float):
...
+        """Return the Beta(1, 1) posterior variance of the ``True`` outcome probability."""
+        alpha, beta = self.posterior_alpha_beta()
+        total = alpha + beta
+        return alpha * beta / (total * total * (total + 1.0))
+
+    def posterior_above(self, threshold: float = 0.5) -> float:
+        """Return the posterior probability that the ``True`` outcome probability
+        exceeds ``threshold`` (default 0.5)."""
+        alpha, beta = self.posterior_alpha_beta()
+        return float(beta_dist.sf(threshold, alpha, beta))
```
ldellant, line 178: where are we using this?

ldellant: it seems enough to me to check that all variances are less than threshold...

src/sympleq/integrations/quantinuum/stitching.py
```diff
@@ -0,0 +1,105 @@
...
+    :param input_circuits: Circuit instances to stitch.
+    :param backend_info: BackendInfo instance containing information on number
+        of device qubits, number of allowed classical register and maximum width
+        for each classical register.
+    :returns: Circuit
+    """
+
+    n_qubits = input_circuits[0].n_qubits
+    if any(n_qubits != c.n_qubits for c in input_circuits[1:]):
+        raise ValueError("All circuits should have the same number of qubits.")
```
ldellant: if we allow to stitch circuits with different qubits here we should simply take the max...

ldellant: ... and then define a new list of circuits, where if the circuit has less qudits than the max value, idle qubits are added...

src/sympleq/integrations/quantinuum/stitching.py
```diff
@@ -0,0 +1,105 @@
...
+        # itself an already-stitched circuit), so wire every one of its bit
+        # registers, in the lexicographic order add_circbox_regwise expects.
+        cregs = []
+        for src_creg in sorted(s_circuit.c_registers, key=lambda r: r.name):
+            cregs.append(sum_circuit.add_c_register(f"creg_{creg_index}", src_creg.size))
+            creg_index += 1
+        sum_circuit.add_circbox_regwise(CircBox(s_circuit), qreg, cregs)
+        if idx == len(input_circuits) - 1:
+            continue
+        sum_circuit.add_circbox(reset_box, sum_circuit.qubits)
```
ldellant: I *think* that to take into ccount the case of differing number of qubits, we should only define the classical registers such that they avoid idle qubits...

tests/core_tests/circuits_tests/circuits_test.py
```diff
@@ -556,7 +556,7 @@ def test_circuit_sanity_checks_exception(self):
             _ = Circuit.from_gates_and_qudits(dimensions, [GATES.CX, GATES.H, GATES.S], [(0, 1), (2,)])
 
     def test_from_random_gates_set_default(self):
-        """Default gate set is {H, S, CX, SWAP}; no ZZPhase or CZ should appear."""
+        """Default gate set is {H, S, CX, SWAP}; no ZZMax or CZ should appear."""
         rng = np.random.default_rng(0)
         c = Circuit.from_random(50, [2, 2, 2], rng=rng)
         allowed = {GATES.H, GATES.S, GATES.CX, GATES.SWAP}
```
ldellant, line 512: For consistency with above, let us have default gate set without `SWAP`

ldellant: If I look at `circuits.py`, line 279, the default gate set contains CX, CZ and SWAP as two-qubits gates...

ldellant: How does this test pass?

src/sympleq/core/circuits/circuits.py
ldellant: Perhaps a good idea to define a default Clifford basis to be shared by all functions? Similar to the default dymension for qudit being equal to 2? 

This way we only call that always and if we change it we are ensured to be consistent everywhere...

tests/core_tests/circuits_tests/circuits_test.py
ldellant: Ah ok makes sense because these are different functions with different Clifford bases... As mentioned earlier, it is probably a good idea to define a Clifford basis and always use that one. For the Clifford basis, only use H, S, and SUM or (generalized) CZ. The basis should work with qudits as well. 

Addendum: We can have two clifford bases: one without paulis (i.e. the one above) and one with paulis (i.e. the one above plus the Paulis X and Z operators)...

tests/core_tests/circuits_tests/circuits_test.py
```diff
@@ -603,6 +603,87 @@ def test_from_random_gates_set_invalid_type(self):
...
+        for g in c.gates:
+            assert g in allowed, f"Default gate set produced unexpected gate {g.name}"
+
+    def test_from_depth_single_qudit_only_covers_full_grid(self):
+        """With ratio=0 every layer is filled with single-qudit gates -> n_gates == depth * n_qudits."""
+        depth = 4
+        dimensions = [2, 2, 2]
+        c = Circuit.from_depth(depth=depth, dimensions=dimensions,
+                               two_qudit_gate_ratio=0.0,
+                               rng=np.random.default_rng(0))
```
ldellant, line 570: is it a good idea to specify the seed here? Aren't we allowing cases in which by chance the seed we have takes works?

tests/core_tests/circuits_tests/gates_test.py
```diff
@@ -321,14 +321,71 @@ def test_gate_from_random_action(self, d: int, n: int):
...
+        """V should equal exp(-iπ/4 X) = (1/√2)(I - iX); V_inv = V†."""
+        U_V = GATES.V.local_unitary(2).toarray()
+        expected = np.array([[1, -1j], [-1j, 1]], dtype=complex) / np.sqrt(2)
+        assert np.allclose(U_V, expected), "V does not equal √X"
+
+        U_V_inv = GATES.V_inv.local_unitary(2).toarray()
+        assert np.allclose(U_V @ U_V_inv, np.eye(2)), "V · V_inv != I"
+
+    def test_V_squared_is_X_up_to_phase(self):
+        """V² = -i·X (up to global phase)."""
```
ldellant: remove "(up to global phase"... If there were a global phase you would get an error...

ldellant (approved):
minor comments really, very well done :)

src/sympleq/core/bayesian_estimation.py
ricott1: as i mentioned, i didnt even check, i just wanted to be sure not to break any workflow for anyone. We can cleanup this file afterwards

src/sympleq/integrations/quantinuum/stitching.py
ricott1: yes, this would be a nice place to do that

src/sympleq/core/circuits/gates.py
ricott1: added a FIXME

src/sympleq/core/bayesian_estimation.py
ricott1: if it's not in probabilities, it means it was not found (yet), and assigning 0 probability is consistent

ldellant (approved):

src/sympleq/core/circuits/circuits.py
ricott1: agree, will need to be fixed later in the application

src/sympleq/integrations/quantinuum/stitching.py
ricott1: stitching with different number of qubits should be part of another PR

src/sympleq/integrations/quantinuum/stitching.py
ricott1: stitching with different number of qubits should be part of another PR

tests/core_tests/circuits_tests/circuits_test.py
ricott1: I agree, the default gate set is a good idea. Please open an issue

src/sympleq/core/circuits/circuits.py
ricott1: Good idea, will do later

RickSimon79221 (approved):
Seems good to me

PR #179, opened 2026-07-06
Stitching checks for pr 178

Shreya93ban:
## 📝 Description
Merge after #178 

- [ ] The code passes CI.


tests/integrations_tests/quantinuum_tests/stitching_diff_qubits_test.py
```diff
@@ -0,0 +1,250 @@
...
+    return tuple(int(x) for x in shots[0])
+
+
+def print_first_commands(circ: Circuit, n: int = 20, title: str = "Circuit"):
+    print(f"\n{title}: first {n} commands")
+    for i, cmd in enumerate(circ.get_commands()[:n]):
+        print(f"{i:03d}: {cmd}")
+
+
+# if __name__ == "__main__":
```
ricott1: please clean up

tests/integrations_tests/quantinuum_tests/stitching_diff_qubits_test.py
```diff
@@ -0,0 +1,250 @@
...
+    print("\n[outcome dict order]")
+    for i, (config, values) in enumerate(outcomes.outcomes.items()):
+        print(
+            f"outcome_index={i:02d} "
+            f"n_qubits={config.n_qubits} "
+            f"n_gates={config.n_gates} "
+            f"outcomes={values}"
+        )
+
+    # print(outcomes)
```
ricott1: clean up

tests/integrations_tests/quantinuum_tests/stitching_diff_qubits_test.py
```diff
@@ -0,0 +1,250 @@
...
+
+#         for failure in failures[:10]:
+#             print(failure)
+
+#         raise AssertionError("Destitched results do not match expected circuit outputs.")
+
+#     print(f"PASSED: all {n_circuits} stitched/destitched circuit results matched.")
+
+
+from sympleq.applications.randomized_benchmarking.config import RMBConfig
```
ricott1: let's put the imports at the top, and clean up duplicates

src/sympleq/integrations/quantinuum/stitching.py
```diff
@@ -63,7 +65,8 @@ def circuit_stitching(
         sum_circuit.add_circbox_regwise(CircBox(s_circuit), qreg, cregs)
         if idx == len(input_circuits) - 1:
             continue
-        sum_circuit.add_circbox(reset_box, s_circuit.qubits)
+        local_reset_box = reset_operations(s_circuit.n_qubits)
+        sum_circuit.add_circbox(local_reset_box, s_circuit.qubits)
```
ricott1, line 70: why we don't reset all qubits? because now with the sorting we have the guarantee that the 'extra' qubits will not be touched anymore?

tests/integrations_tests/quantinuum_tests/stitching_diff_qubits_test.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+
+from pytket.circuit import Circuit
+
+from sympleq.integrations.quantinuum.stitching import circuit_stitching
+
+def _deterministic_native_circuit(n_qubits: int, seed: int) -> tuple[Circuit, tuple[int, ...]]:
+    """Build a native circuit with a known computational-basis output."""
+    circuit = Circuit(n_qubits, n_qubits)
+    expected = tuple((seed + qubit) % 2 for qubit in range(n_qubits))
```
ricott1: so this is just a tuple of alternating 0 and 1s, starting with either 0 or 1?

src/sympleq/integrations/quantinuum/stitching.py
ldellant: we do not need to reset the ones we do not use... (?)

Shreya93ban: I was doing this, actually two commits back it exactly had this resetting all qubit :).  Can revert back to it, however, this is more general, I think?

tests/integrations_tests/quantinuum_tests/stitching_diff_qubits_test.py
Shreya93ban: This is only relevant if we are actually checking the answers, which we are not in this test.  This was relevant when I was manually checking the outcomes. The deterministic circuit is designed to have this output

src/sympleq/integrations/quantinuum/stitching.py
ricott1: No need to revert, i just wanted to clarify the reason (maybe a comment would also be useful here)

tests/integrations_tests/quantinuum_tests/stitching_diff_qubits_test.py
Shreya93ban: I got rid of the expected outcomes, as this is not needed for this test

src/sympleq/integrations/quantinuum/stitching.py
Shreya93ban: Do you mean a comment in the file? I added it, please let me know if this is what you mean.

Shreya93ban:
Requires #198 to be fixed first

PR #186, opened 2026-07-17
Split core Clifford symplectic utilities

c-nation:
## 📝 Description
Adds reusable core Clifford/symplectic utilities for lifting Pauli tableau mappings to Clifford/symplectic maps.
Main changes:
Adds inverse-based Pauli tableau mapping as the default path, with transvection version as option.
Adds permutation-to-symplectic/Clifford utilities in core, independent of graph automorphism search.
Extends qudit-aware finite-field solving and symplectic mapping.
Updates gate/decomposition call sites and focused tests for the new mapping API.

- [x] The code passes CI.


src/sympleq/core/circuits/find_symplectic.py
ldellant, line 1: For a future PR (leave a `#FIXME` in the code)

In the long term, I do not see a reason to distinguish between the `qubit` and `qudit` (next file). If there is, let's chat about it; otherwise let us have a new PR (after the `dev` version is again cleaned) where all functions to find the symplectic are efficient and (mixed) qudit ready.

ldellant, line 1: For a future PR (leave a `#FIXME` in the code)

It seems that the logic can be simplified quite a bit here... I would use `S = input^-1 target` (sorry if typos) plus some check before that ensure consistency... This should be applicable to qudits as well and should be quite efficient? 

Then it is only a matter to fix the phase, which should be always (? please correct me if wrong) doable efficiently...

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
ldellant, line 1: For a future PR (put a `#FIXME:` in the code)

this is a very nice file, that does something important to us. I think that at this moment:

- is not very clear
- needs to be generalized to different `gate_set`. This is important for expressing the circuit with only gates that are native to the platform we are using (e.g. quantinuum)

RickSimon79221 (approved):

src/sympleq/utils.py
```diff
@@ -110,4 +110,32 @@ def complex_phase_value(phase: int, dimension: int) -> complex:
...
+    """
+    Compute the Kronecker product of multiple matrices.
+
+    Args:
+        matrices (List[np.ndarray]): A list of matrices to compute the Kronecker product of.
+
+    Returns:
+        np.ndarray: The Kronecker product of the input matrices.
+    """
+    if len(matrices) <= 1:
```
ricott1: this breaks for empty `matrices`

src/sympleq/utils.py
```diff
@@ -110,4 +110,32 @@ def complex_phase_value(phase: int, dimension: int) -> complex:
...
+
+    # Avoid roundoff for the common quadrant roots 1, i, -1, -i.
+    if (2 * phase) % dimension == 0:
+        quadrant = ((2 * phase) // dimension) % 4
+        return (1.0 + 0.0j, 0.0 + 1.0j, -1.0 + 0.0j, 0.0 - 1.0j)[quadrant]
+
     return np.exp(2 * np.pi * 1j * phase / (2 * dimension))
+
+
+def multi_kron(matrices):
```
ricott1: is this in the hot path? cause maybe smtg like reduce would avoid the python loop

src/sympleq/core/circuits/gates.py
```diff
@@ -145,12 +145,12 @@ def from_random(
         return _GenericGate("random", symplectic, phase_vector)
 
     @classmethod
-    def solve_from_target(cls, input_tableau: TableauLike, target_tableau: TableauLike) -> Gate:
+    def solve_from_target(cls, input_tableau: TableauLike, target_tableau: TableauLike, dimension: int) -> Gate:
```
ricott1: could we have a default value for dimension so we dont break the API?

src/sympleq/core/circuits/find_symplectic.py
```diff
@@ -35,14 +48,14 @@ def find_symplectic_solution(u: IntNDArray, v: IntNDArray) -> IntNDArray:
...
         return direct_construction(u, v)
-    else:
-        # u and v are symplectically orthogonal - use general linear system solver
-        return solve_general_system(u, v)
+
+    # General case over GF(p): solve linear constraints directly.
+    return solve_general_system(u, v, p=p)
 
 
 def direct_construction(u: IntNDArray, v: IntNDArray) -> IntNDArray:
```
ricott1, line 69: why dont we add the `p` argument here?

src/sympleq/core/finite_field_solvers.py
```diff
@@ -508,6 +546,52 @@ def gf_inv(A, p: int = 2):
     return AI[:, n:] % p
 
 
+def gf2_inv(M: np.ndarray) -> np.ndarray:
```
ricott1: We should use the standard typing we use in the rest of the code, e.g. IntNDArray

src/sympleq/core/circuits/find_symplectic.py
cerisola: I think this method only works for the p=2 case.

src/sympleq/core/circuits/gates.py
ldellant: For a future PR, this function should work for (mixed) prime qudits. 

Add a `FIXME:` Generalize this function to work with mixed prime qudits

src/sympleq/core/circuits/find_symplectic_qudits.py
ldellant, line 1: For a future PR (put a #FIXME: in the code)

Eliminate this file and generalize functions in `find_symplectic.py` to work with (mixed) prime dimensions

src/sympleq/core/circuits/target.py
ldellant, line 1: For a future PR (leave a `#FIXME` in the code)

This is part of the files that determine the gate from an input PuliSum to an output PauliSum. I think we should have all these functions in a single file, something called `input_to_target.py`, that contains **everything** related to this feature of the code. Then, a good idea would be for `gates.py` to import relevant functions and add a method there that creates a gate object that from an input PauliSum yields a target PauliSum.

src/sympleq/core/finite_field_solvers.py
ldellant, line 1: This file seems a bit misplaced... Is it useful only for finding a symplectic from input to target?

tests/core_tests/circuits_tests/gates_test.py
```diff
@@ -304,7 +304,8 @@ def test_gate_from_target(self):
         # Map X to Z on a single qubit: [1, 0] -> [0, 1]
         input_tableau = np.array([[1, 0]])
         target_tableau = np.array([[0, 1]])
-        gate = Gate.solve_from_target(input_tableau, target_tableau)
+        dim = 2
```
ldellant, line 307: want to generalise to different dimensions in the future (mixed as well)

src/sympleq/utils.py
ldellant: should have solved it :)

src/sympleq/utils.py
ldellant: I think it is not necessary as the number of iteration is very much not the limiting factor (that being the kron operation)...

ricott1: add typing and return type

src/sympleq/utils.py
ricott1: what you mean? I think it's still broken

src/sympleq/utils.py
ricott1: sounds good

src/sympleq/core/circuits/find_symplectic.py
```diff
@@ -108,64 +129,71 @@ def direct_construction(u: IntNDArray, v: IntNDArray) -> IntNDArray:
...
     Returns:
-        Binary vector w of length 2n such that <u,w> = <v,w> = 1.
+        Vector w of length 2n such that <u,w> = <v,w> = 1.
 
     Raises:
-        Exception: If no solution exists in GF(2).
+        Exception: If no solution exists in GF(p).
     """
-    n = len(u) // 2
+    n = len(u) // 2  # kept for shape clarity
```
ricott1: remove this stranded comment?

src/sympleq/utils.py
```diff
@@ -110,4 +114,44 @@ def complex_phase_value(phase: int, dimension: int) -> complex:
...
+        return 1j ** quadrant
+    else:
+        return np.exp(2 * np.pi * 1j * phase / (2 * dimension))
+
+
+def multi_kron(matrices: Sequence[ComplexNDArray]) -> ComplexNDArray:
+    """
+    Compute the Kronecker product of multiple matrices.
+
+    Args:
```
ricott1: let s use numpy style docstrings

src/sympleq/core/circuits/find_symplectic.py
```diff
@@ -238,56 +272,189 @@ def check_mappable_via_clifford(pauli_sum_tableau: TableauType,
...
+    pauli_sum_tableau: TableauType,
+    target_pauli_sum_tableau: TableauType,
+    p: int = 2,
+    method: str = "inverse",
+) -> TableauType:
     """
-    Map a Pauli sum to a target Pauli sum using symplectic transvections.
+    Map a Pauli sum tableau to a target tableau.
+
+    ``method="inverse"`` is the default and computes the unique row-action map
```
ricott1, line 337: This to me looks a bit more complex than necessary. Let's add a FIXME to clean it up in the future (we probably could get away with a bool rather than a string method)

src/sympleq/core/circuits/find_symplectic.py
```diff
@@ -238,56 +272,189 @@ def check_mappable_via_clifford(pauli_sum_tableau: TableauType,
...
+            )
+
+    return _map_pauli_sum_to_target_tableau_by_transvections(
+        pauli_sum_tableau,
+        target_pauli_sum_tableau,
+        p=p,
+    )
+
+
+def _map_pauli_sum_to_target_tableau_by_transvections(
```
ricott1: let's be consistent with pauli_sum vs paulisum

src/sympleq/core/circuits/find_symplectic.py
```diff
@@ -238,56 +272,189 @@ def check_mappable_via_clifford(pauli_sum_tableau: TableauType,
...
+    """
+    Return the row-action symplectic map induced by a Pauli-row permutation.
+
+    The returned matrix ``F`` satisfies ``tableau @ F == tableau[permutation]``
+    modulo ``p`` when the permutation preserves the symplectic product matrix.
+    This is graph-independent core logic used by higher-level symmetry finders.
+    The default method is the inverse-basis lift.
+    """
+    tableau = np.asarray(pauli_sum_tableau, dtype=int) % p
+    pi = np.asarray(permutation, dtype=np.int64).reshape(-1)
```
ricott1, line 413: id use dtype=int everywhere

src/sympleq/core/circuits/find_symplectic.py
```diff
@@ -238,56 +272,189 @@ def check_mappable_via_clifford(pauli_sum_tableau: TableauType,
...
+            F = np.asarray(F, dtype=int)
+        else:
+            GF = galois.GF(int(p))
+            F_gf = np.linalg.inv(GF(input_basis)) @ GF(output_basis)
+            F = np.asarray(F_gf, dtype=int) % p
+    except np.linalg.LinAlgError:
+        return None
+
+    if not np.array_equal((input_tab @ F) % p, output_tab):
+        raise ValueError("Complete-basis map failed to reproduce the target tableau.")
```
ricott1, line 465: should we raise or return None?

src/sympleq/core/circuits/gate_decomposition_to_circuit.py
```diff
@@ -83,7 +91,26 @@ def _is_invertible_mod(A: np.ndarray, p: int) -> bool:
...
+        return None
+    return None
+
+
+def ensure_invertible_A_circuit(
+    F: TableauType,
+    p: int,
+    max_depth: int | None = None,
+    *,
+    max_bfs_states: int | None = 1_000_000,
```
ricott1: if possible, remove this stuff about memory

src/sympleq/core/circuits/find_symplectic.py
```diff
@@ -238,56 +272,189 @@ def check_mappable_via_clifford(pauli_sum_tableau: TableauType,
...
+    try:
+        if p == 2:
+            input_basis_inv = gf2_inv(input_basis)
+            F = (input_basis_inv @ output_basis) & 1
+            F = np.asarray(F, dtype=int)
+        else:
+            GF = galois.GF(int(p))
+            F_gf = np.linalg.inv(GF(input_basis)) @ GF(output_basis)
+            F = np.asarray(F_gf, dtype=int) % p
+    except np.linalg.LinAlgError:
```
ricott1, line 461: then maybe except np.linalg.LinAlgError as e, and raise(e)?

src/sympleq/core/circuits/find_symplectic.py
ldellant: Left a comment explaining how it should behave

src/sympleq/core/circuits/find_symplectic.py
ldellant: again left a comment on how this should behave.

src/sympleq/core/circuits/find_symplectic.py
ldellant: I agree yet this is a problem everywhere, not only for this PR... We should have a new PR only for this

ricott1 (approved):

PR #188, opened 2026-07-17
Split/core minimal qudit frame

c-nation:
## 📝 Description

Third one to do in order. Biggest addition - code for the minimal decomposition and testing.

- [ ] The code passes CI.


RickSimon79221 (approved):
Obviously not able to check that everything is correct, but I am fairly certain, that nothing here will break dev

PR #189, opened 2026-07-17
Split/models

c-nation:
## 📝 Description

4th PR to do in order...
Update to models and tests for them

- [ ] The code passes CI.


RickSimon79221 (approved):
PR mainly introduces new models and tests for the models. No real threat to cause problems in dev. Only thing I could see that might have impact, is the change of random generator for symmetric hamiltonians. But since it still works with all the tests I am not really worried

PR #194, opened 2026-07-29
Fix mixed dimension case for `PauliString.acquired_phase`

ricott1:
Add fix for mixed dimension also for PauliString.acquired_phase

## 📝 Description

- [ ] The code passes CI.

### DIFF
--- a/tests/core_tests/paulis_tests/pauli_test.py
+++ b/tests/core_tests/paulis_tests/pauli_test.py
@@ -462,6 +462,25 @@ class TestPaulis:
             f"with dimensions {dimensions}"
         )
 
+    def test_acquired_phase_on_mixed_dimensions(self):
+        dims = [2, 3]
+        lcm = 6
+
+        z0 = PauliString.from_string('x0z1 x0z0', dimensions=dims)
+        x0 = PauliString.from_string('x1z0 x0z0', dimensions=dims)
+        z1 = PauliString.from_string('x0z0 x0z1', dimensions=dims)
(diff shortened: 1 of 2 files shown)


RickSimon79221 (approved):

tests/core_tests/paulis_tests/pauli_test.py
```diff
@@ -462,6 +462,25 @@ def test_pauli_sum_addition(self):
...
+        assert z0.acquired_phase(x0) == 2 * (lcm // dims[0])
+        assert z1.acquired_phase(x1) == 2 * (lcm // dims[1])
+        assert x0.acquired_phase(z0) == 0
+        assert x1.acquired_phase(z1) == 0
+
+        for a, b in [(z0, x0), (z1, x1), (x0, z0), (x1, z1)]:
+            assert a.acquired_phase(b) == (a * b).phases[0] % (2 * lcm), (
+                f'acquired_phase disagrees with __mul__ for {a} and {b}'
+            )
+
```
ldellant, line 483: I think this test is a bit self-referencing, in the sense that it asks whether the logic used to define the `__mul__` is the logic used to define it. This is absolutely OK, and we should not change that, yet we should add another assert in the loop below such that:

```
matrix_from_a = a.to_hilbert_space()
matrix_from_b = b.to_hilbert_space()
matrix_from_ab = (a*b).to_hilbert_space()

assert matrix_from_a @ matrix_from_b == matrix_from_ab
```

c-nation (approved):

PR #196, opened 2026-08-05
Solve_from_target

Shreya93ban:
## 📝 Description
 Clean up for uniform dimension qudits, focused on 'solve_from_target'

- [ ] The code passes CI.


src/sympleq/core/circuits/helpers_from_input_to_target.py
```diff
@@ -0,0 +1,295 @@
...
+
+    if len(basis_indices) < n_cols:
+        input_basis, output_basis = complete_basis(input_tab, output_tab, p)
+
+    else:
+        input_basis = input_tab[basis_indices]
+        output_basis = output_tab[basis_indices]
+
+    try:
+        F = (gf_inv(input_basis, p) @ output_basis) % p
```
c-nation, line 163: Is this the correct way around for the current act? We have p F^T in the code right? (perhaps we want to alter to p F long term?)

src/sympleq/core/circuits/gates.py
```diff
@@ -6,16 +6,20 @@
...
     symplectic_random_koenig_smolin_gf2,
     symplectic_random_transvection,
 )
-from sympleq.core.circuits.find_symplectic import map_paulisum_to_target_tableau
+from sympleq.core.circuits.helpers_solve_from_target import (
+    check_mappable_via_clifford,
+    get_phase_vector,
+    map_tableau_to_target_tableau,
+    solve_phase_for_gate,
+)
```
ldellant: in the long term we would like all functions that are required to `solve_from_target` into a single file. Is that file `helpers_solve_from_target`? If yes, perhaps let's change the name of that file (and of the function calling it) to be `helpers_from_input_to_output` and `from_input_to_output`. Other names are also OK, but I think that `solve_from_target` is misleading as it is suggesting we only need the target, while we also need the input :)

src/sympleq/core/circuits/helpers_from_input_to_target.py
```diff
@@ -0,0 +1,295 @@
...
+    h_p = solve_linear_system_over_gf(A % p, delta % p, p)
+    # Solve mod 2
+    h_2 = solve_linear_system_over_gf(A % 2, delta % 2, 2)
+
+    h_p, h_2 = np.asarray(h_p, int) % p, np.asarray(h_2, int) % 2
+    return (h_p + p * ((h_2 - h_p) % 2)) % (2 * p)
+
+
+# To be removed??
+# Could be useful for mixed??
```
ldellant: these comments are a bit strange... It seems that the function `get_phase_vector` is not used? If this is the case, perhaps do a PR (or a commit) just to remove it, such that it is easy to find it again. We should not have two functions that do the same thing (as it seems to be the case below)

src/sympleq/core/circuits/target.py
ldellant, line 1: Can we move all of this file into `helpers_solve_from_target`?

src/sympleq/core/phase_correction/solve.py
```diff
@@ -63,6 +63,7 @@ def pauli_phase_correction(H: TableauType, delta_phi_2p: PhasesType, p: int, dim
     return PauliGate(pauli)
 
 
+# Can be cleaned up
```
ldellant: if you are leaving the comment, add a `# TODO:`

src/sympleq/core/phase_correction/solve.py
```diff
@@ -176,6 +177,7 @@ def solve_phase_vector_h_from_residual(
     return None
 
 
+# Shreya: This can be removed I think. We already have something similar in sympleq.core.finite_field_solvers
```
ldellant, line 180: Totally agree, and we should really remove redundancies. This is for a new PR, if you think you can forget it please add an issue.

src/sympleq/core/phase_correction/solve.py
ldellant, line 1: Do we need an entire folder for the phase correction? This may be a broader conversation between us. If yes, move everything related to phase correction here.

tests/core_tests/circuits_tests/helpers_from_input_to_target_test.py
ldellant, line 1: use the same name convention 

`helpers_solve_from_target_test`

ldellant (changes requested):
Follow the codecov coverage (in Conversation) to see which lines are not tested and insert tests for those.

tests/core_tests/circuits_tests/helpers_from_input_to_target_test.py
Shreya93ban: as the name of the file changed, changed the name to helpers_from_input_to_target.test

src/sympleq/core/phase_correction/solve.py
Shreya93ban: Added an issue #201

src/sympleq/core/circuits/helpers_from_input_to_target.py
Shreya93ban: Added a TODO for a future PR, the function is used. It is slightly different than the function below; which requires a target PauliSum. The get_phase_vector takes in just the input and dimension and provides a seed for the phase-solver, which is non-trivial for qubits. So based on where this function is used, either the function below needs to be modified, or this can be moved to another location.

